Issue #17, opened 2024-10-31
[Bug] Mellinger controller stops working correctly when yaw is not close to zero


j-kelm:
When doing the following:
- starting drone with yaw = 0 and hovering
- rotating yaw set point to e.g. pi/2
- applying minor disturbances in roll or pitch
the emulated Mellinger controller fails to stabilize the drone. 

This behavior does not occur on the real drone and is related to the state and gyro values given to the emulated Mellinger controller (which would be obtained from the sensors on the real drone).

The fix involves using a different angle rotation order for providing the attitude to the Mellinger controller as well as rotating the angular velocity into the body frame (which did not happen this far).

amacati:
I have looked at the PR and replicated the issue. Since I can't make changes to your PR I took the liberty to fix it on dev while adding tests and fixing other related issues in the physics module.

PR #18, opened 2024-10-31
Fixed bug introduced by supplying wrong state to Mellinger

j-kelm:
Fix for #17 

Changed intrinsic XZY to extrinsic xyz for rotations, used ang_vel from sim instead of finite differences for Mellinger, transformed ang_vel from world to body for Mellinger,


amacati:
I wanted to have automated testing for this issue and then realized that the dyn physics mode was also broken because of this issue. Since I could not edit this PR I have fixed the original issue in 1ff2931 while also adding the fixes for the physics mode and more tests. You are now co-author of that commit instead, I hope that's okay with you.

PR #23, opened 2024-12-24
Removed closing "controller" and replaced it with crazyswarm magic

### LINKED ISSUE #22: Closing controller doesn't consider yaw
The closing controller commands yaw=0, but the drone might pass the gate with different yaw. When returning to the starting position, the yaw should again be 0.

j-kelm:
Merry Christmas, everyone! Rejoice, Santa has finally brought a working™ closing "controller" so that the amount of innocent drone casualties may be reduced in the upcoming year.

Since it uses crazyswarm directly, it is incompatible with the existing closing controller architecture. Thus, my changes include fully removing the last iteration of the closing controller.

Thoroughly tested on hardware for varying speeds and race tracks. A little hacky, but works. Also fixes #22.


lsy_drone_racing/envs/drone_racing_deploy_env.py
```diff
@@ -179,45 +178,29 @@ def step(
...
+            self.cf.goTo(goal = return_pos, yaw=0, duration=BREAKING_DURATION)
+            time.sleep(BREAKING_DURATION - 1)
+
+            return_pos[:2] = self.config.env.track.drone.pos[:2]
+            self.cf.goTo(goal=return_pos, yaw=0, duration=RETURN_DURATION)
+            time.sleep(RETURN_DURATION)
+
+            self.cf.land(self.config.env.track.drone.pos[2], LAND_DURATION)
+            time.sleep(LAND_DURATION)
+        except Exception as e:
```
amacati, line 202: Do you need the general Exception catch here? If so, why?

lsy_drone_racing/envs/drone_racing_deploy_env.py
```diff
@@ -179,45 +178,29 @@ def step(
...
-                controller.step_callback(action, obs, 0, True, False, info)
-                time.sleep(t_step_ctrl)
-
-        self.cf.notifySetpointsStop()
-        self.cf.land(0.05, 2.0)
+            
+            return_pos = obs['pos'] + obs['vel']/(np.linalg.norm(obs['vel']) + 1e-8) * BREAKING_DISTANCE
+            return_pos[2] = RETURN_HEIGHT
+            self.cf.goTo(goal = return_pos, yaw=0, duration=BREAKING_DURATION)
+            time.sleep(BREAKING_DURATION - 1)
```
amacati, line 194:
```suggestion
            time.sleep(BREAKING_DURATION - 1)  # Smoothly transition to next position by setting the next goal earlier
```

amacati (changes requested):
Looks good. Can you address the minor comments I had? Then we can merge and save a few drone markers from a premature demise.

lsy_drone_racing/envs/drone_racing_deploy_env.py
j-kelm: The general Exception catch is not necessary and admittedly bad code. At least notifysetpointstop and maybe also the other calls to cf may raise a rospy.service.ServiceException if the drone is not reachable (e.g. because of an emergency shutoff of the drone), resulting in hanging code in my case.

Catching the rospy.service.ServiceException should probably be enough, but I cannot validate this on the drone anymore.

PR #55, opened 2025-11-07
Fix track randomization - Fixes #53

### LINKED ISSUE #53: [Bug] Tracks are not regenerated across runs in level 3
It is reported by the students that the nominal layouts are not rearranged across runs in level 3.

niklasschlueter:


lsy_drone_racing/envs/race_core.py
```diff
@@ -325,14 +325,8 @@ def __init__(
...
-        self.randomize_track = build_track_randomization_fn(
-            randomizations,
-            gate_ids,
-            obstacle_ids,
-            self.gates["nominal_pos"],
-            self.gates["nominal_quat"],
-            self.obstacles["nominal_pos"],
-        )
+        self.randomize_track = build_track_randomization_fn(randomizations, gate_ids, obstacle_ids)
+        self.track_gen_key = jax.random.PRNGKey(seed)
```
amacati: Why a dedicated key? We have one already from the sim

lsy_drone_racing/envs/race_core.py
```diff
@@ -354,9 +348,27 @@ def _reset(
         # explicitly do it here
         self.sim.reset(mask=mask)
         key, subkey = jax.random.split(self.sim.data.core.rng_key)
+        # Generate random track
+        self.track_gen_key, track_sub_key = jax.random.split(self.track_gen_key)
```
amacati: You can split multiple keys from jax.random.split. It's more efficient to just generate three instead of two one call further up

lsy_drone_racing/envs/race_core.py
```diff
@@ -354,9 +348,27 @@ def _reset(
...
+        track = (
+            generate_random_track(self.track, track_sub_key) if self.track.randomize else self.track
+        )
+        self.gates, self.obstacles, self.drone = load_track(track)
         # Randomize the track
         self.sim.data = self.sim.data.replace(core=self.sim.data.core.replace(rng_key=key))
-        self.sim.mjx_data = self.randomize_track(self.sim.mjx_data, mask, subkey)
+
+        # Randomized drone pos
+        pos = self.sim.data.states.pos.at[...].set(self.drone["pos"])
```
amacati: It might make sense to jit this and the following lines

PR #56, opened 2025-11-17
Added Level-3 support for deployment & miscs

N0OBSTUDENT:
-The real race environment could read the actual track layout and "randomize" the layout from MoCapas nominal track info
-Refactored checks.py, making it independent to `ROSConnector`. The only file that calls it will be real_race_env.py.
-Reorganized real race environment reset procedure. Track objects are checked before the drone connects. Now the students do not need to wait a long time for drone connection before they  know whether they are placing the track objects at the correct position
-Now with `--save_track some_config_name.toml` option while calling `python deploy.py`, the actual track layout with other the configuration from the original config file will be stored as backup in `config/some_config_name.toml`, so that the students could check setup after the real-world run.

### DIFF
--- a/lsy_drone_racing/envs/real_race_env.py
+++ b/lsy_drone_racing/envs/real_race_env.py
@@ -48,6 +48,9 @@ class EnvData:
     gates_visited: NDArray
     obstacles_visited: NDArray
     last_drone_pos: NDArray[np.float32]
+    taken_off: bool = False
+    drone_connected: bool = False
+
 
     @classmethod
     def create(cls, n_drones: int, n_gates: int, n_obstacles: int) -> EnvData:
@@ -58,6 +61,15 @@ class EnvData:
             obstacles_visited=np.zeros((n_drones, n_obstacles), dtype=bool),
             last_drone_pos=np.zeros((n_drones, 3), dtype=np.float32),
         )
+
+    def reset(self, last_drone_pos: NDArray[np.float32]):
+        """Reset the environment data."""
+        self.target_gate[...] = 0
+        self.gates_visited[...] = False
+        self.obstacles_visited[...] = False
+        self.last_drone_pos[...] = last_drone_pos
+        self.taken_off = False
+        self.drone_connected = False
 
 
 # region CoreEnv
@@ -111,7 +123,6 @@ class RealRaceCoreEnv:
         assert control_mode in ["state", "attitude"], f"Invalid control mode {control_mode}"
         self.control_mode = control_mode
         self.randomizations = randomizations
-        self._connected_to_drone = False
         self.drone_parameters = load_params("first_principles", drones[rank]["drone_model"])
         self.drone = Crazyflie(rw_cache=str(Path(__file__).parent / ".cache"))
         self._drone_healthy = mp.Event()
@@ -121,9 +132,6 @@ class RealRaceCoreEnv:
             cmd_topic=f"/drones/{self.drone_name}/command",
             timeout=10.0,
         )
-        # A hatched indicator whether the drone has taken off at least once
-        # Used to prevent the return controller from engaging before takeoff
-        self._taken_off = False
         # Dynamic data
         self.data = EnvData.create(
             n_drones=self.n_drones, n_gates=self.n_gates, n_obstacles=self.n_obstacles
@@ -132,23 +140,25 @@ class RealRaceCoreEnv:
 
     def _reset(self, *, seed: int | None = None, options: dict | None = None) -> tuple[dict, dict]:
         """Reset the environment and return the initial observation and info."""
-        real_track = options.get("real_track_objects", True)
-        check_track = options.get("check_race_track", True)
-        check_drone = options.get("check_drone_start_pos", True)
-
-        if real_track:
+        options = {} if options is None else options
+        # Update the position of gates and obstacles with the real positions measured from Mocap. If
+        # disabled, they are equal to the nominal positions defined in the track config.
+        if options.get("real_track_objects", True):
             self._update_track_poses()
-            if check_track:
-                check_race_track(
-                    gates=self.gates, obstacles=self.obstacles, rng_config=self.randomizations
-                )
-
-        if check_drone:
-            real_pos = self._ros_connector.pos[self.drone_name]
-            nominal_pos = np.array(self.drones.pos[self.rank])
+        if options.get("check_race_track", True):
+            check_race_track(
+                gates_pos=self.gates.pos,
+                nominal_gates_pos=self.gates.nominal_pos,
+                gates_quat=self.gates.quat,
+                nominal_gates_quat=self.gates.nominal_quat,
+                obstacles_pos=self.obstacles.pos,
+                nominal_obstacles_pos=self.obstacles.nominal_pos,
+                rng_config=self.randomizations,
+            )
+        if options.get("check_drone_start_pos", True):
             check_drone_start_pos(
-                nominal_pos=nominal_pos,
-                real_pos=real_pos,
+                nominal_pos=self.drones.pos[self.rank],
+                real_pos=self._ros_connector.pos[self.drone_name],
                 rng_config=self.randomizations,
                 drone_name=self.drone_name,
             )
@@ -157,7 +167,7 @@ class RealRaceCoreEnv:
             radio_id=self.rank, radio_channel=self.drone_channel, drone_id=self.drone_id
         )
         self._last_drone_pos_update = 0  # Last time a position was sent to the drone estimator
-        self._reset_env_data(self.data)
+        self.data.reset(np.stack([self._ros_connector.pos[n] for n in self.drone_names]))
         self._reset_drone()
 
         if self.control_mode == "attitude":
@@ -191,7 +201,7 @@ class RealRaceCoreEnv:
         self.data.target_gate += np.asarray(passed)
         self.data.target_gate[self.data.target_gate >= self.n_gates] = -1
         self.data.last_drone_pos[...] = drone_pos
-        self._taken_off |= drone_pos[self.rank, 2] > 0.2
+        self.data.taken_off |= drone_pos[self.rank, 2] > 0.2
         # Send vicon position updates to the drone at a fixed frequency irrespective of the env freq
         # Sending too many updates may deteriorate the performance of the drone, hence the limiter
         if (t := time.perf_counter()) - self._last_drone_pos_update > 1 / self.POS_UPDATE_FREQ:
@@ -301,16 +311,16 @@ class RealRaceCoreEnv:
             for i in range(self.n_obstacles):
                 self.obstacles.pos[i, ...] = pos[f"obstacle{i + 1}"]
         except KeyError as e:
-            raise ValueError(
-                f"Could not find all track objects in the ROS TF tree: {e}. \
-                            Have you enabled the track objects in Vicon/ \
-                            started the motion capture tracking node?"
+            raise KeyError(
+                f"Could not find all track objects in the ROS TF tree: {e}. Have you enabled the "
+                "track objects in Vicon and started the motion capture tracking node?"
             ) from e
 
     def _connect_radio(self, radio_id: int, radio_channel: int, drone_id: int):
         """Connect to the drone via radio.
 
-        If the drone is not reachable, a TimeoutError is raised.
+        Raises:
+            TimeoutError: If the drone is not reachable within 10 seconds.
         """
         cflib.crtp.init_drivers()
         uri = f"radio://{radio_id}/{radio_channel}/2M/E7E7E7E7" + f"{drone_id:02x}".upper()
@@ -339,16 +349,8 @@ class RealRaceCoreEnv:
         connected = event.wait(10.0)
         if not connected:
             raise TimeoutError("Timed out while waiting for the drone.")
-        self._connected_to_drone = True
+        self.data.drone_connected = True
         logger.info(f"Drone {drone_id} connected to {uri}")
-
-    def _reset_env_data(self, data: EnvData):
-        """Reset the environment data."""
-        data.target_gate[...] = 0
-        data.gates_visited[...] = False
-        data.obstacles_visited[...] = False
-        drone_pos = np.stack([self._ros_connector.pos[n] for n in self.drone_names])
-        data.last_drone_pos[...] = drone_pos
 
     def _reset_drone(self):
         """Arm the drone, reset estimation."""
@@ -447,19 +449,20 @@ class RealRaceCoreEnv:
         Irrespective of succeeding or not, the drone will be stopped immediately afterwards or in
         case of errors, and close the connections to the ROSConnector.
         """
+        if not self.data.drone_connected or not self.data.taken_off:
+            self._ros_connector.close()
+            return
         try:
-            if self._connected_to_drone and self._taken_off:
-                self._return_to_start()
+            self._return_to_start()
         finally:
             try:
-                if self._connected_to_drone:
-                    # Kill the drone
-                    pk = CRTPPacket()
-                    pk.port = CRTPPort.LOCALIZATION
-                    pk.channel = Localization.GENERIC_CH
-                    pk.data = struct.pack("<B", Localization.EMERGENCY_STOP)
-                    self.drone.send_packet(pk)
-                    self.drone.close_link()
+                # Kill the drone
+                pk = CRTPPacket()
+                pk.port = CRTPPort.LOCALIZATION
+                pk.channel = Localization.GENERIC_CH
+                pk.data = struct.pack("<B", Localization.EMERGENCY_STOP)
+                self.drone.send_packet(pk)
+                self.drone.close_link()
             finally:
                 # Close all ROS connections
                 self._ros_connector.close()
--- a/lsy_drone_racing/envs/real_race_env.py
+++ b/lsy_drone_racing/envs/real_race_env.py
@@ -27,17 +27,13 @@ from drone_estimators.ros_nodes.ros2_connector import ROSConnector
 from drone_models.core import load_params
 from drone_models.transform import force2pwm
 from gymnasium import Env
-from ml_collections import ConfigDict
 from scipy.spatial.transform import Rotation as R
 
-from lsy_drone_racing.envs.utils import gate_passed, load_track, randomize_track
-from lsy_drone_racing.utils.checks import (
-    check_drone_start_pos,
-    check_gates_in_bound,
-    check_race_track,
-)
+from lsy_drone_racing.envs.utils import gate_passed, load_track
+from lsy_drone_racing.utils.checks import check_drone_start_pos, check_race_track
 
 if TYPE_CHECKING:
+    from ml_collections import ConfigDict
     from numpy.typing import NDArray
 
 logger = logging.getLogger(__name__)
@@ -51,6 +47,7 @@ class EnvData:
     gates_visited: NDArray
     obstacles_visited: NDArray
     last_drone_pos: NDArray[np.float32]
+    taken_off: bool = False
 
     @classmethod
     def create(cls, n_drones: int, n_gates: int, n_obstacles: int) -> EnvData:
@@ -61,6 +58,15 @@ class EnvData:
             obstacles_visited=np.zeros((n_drones, n_obstacles), dtype=bool),
             last_drone_pos=np.zeros((n_drones, 3), dtype=np.float32),
         )
+
+    def reset(self, last_drone_pos: NDArray[np.float32]):
+        """Reset the environment data."""
+        self.target_gate[...] = 0
+        self.gates_visited[...] = False
+        self.obstacles_visited[...] = False
+        self.last_drone_pos[...] = last_drone_pos
+        self.taken_off = False
+        self.drone_connected = False
 
 
 # region CoreEnv
@@ -113,9 +119,7 @@ class RealRaceCoreEnv:
         self.device = jax.devices("cpu")[0]
         assert control_mode in ["state", "attitude"], f"Invalid control mode {control_mode}"
         self.control_mode = control_mode
-        self.predefined_track = not (track.get("randomize", False))
         self.randomizations = randomizations
-        self._connected_to_drone = False
         self.drone_parameters = load_params("first_principles", drones[rank]["drone_model"])
         self.drone = Crazyflie(rw_cache=str(Path(__file__).parent / ".cache"))
         self._drone_healthy = mp.Event()
@@ -125,9 +129,6 @@ class RealRaceCoreEnv:
             cmd_topic=f"/drones/{self.drone_name}/command",
             timeout=10.0,
         )
-        # A hatched indicator whether the drone has taken off at least once
-        # Used to prevent the return controller from engaging before takeoff
-        self._taken_off = False
         # Dynamic data
         self.data = EnvData.create(
             n_drones=self.n_drones, n_gates=self.n_gates, n_obstacles=self.n_obstacles
@@ -136,45 +137,25 @@ class RealRaceCoreEnv:
 
     def _reset(self, *, seed: int | None = None, options: dict | None = None) -> tuple[dict, dict]:
         """Reset the environment and return the initial observation and info."""
-        real_track = options.get("real_track_objects", True)
-        check_track = options.get("check_race_track", True)
-        check_drone = options.get("check_drone_start_pos", True)
-
-        if (not self.predefined_track) and (not real_track):
-            raise ValueError(
-                "Randomly generated track requires real_track_objects "
-                "to be True to get the actual track poses."
-            )
-
-        if not self.predefined_track:
+        options = {} if options is None else options
+        # Update the position of gates and obstacles with the real positions measured from Mocap. If
+        # disabled, they are equal to the nominal positions defined in the track config.
+        if options.get("real_track_objects", True):
             self._update_track_poses()
-            self._update_drone_starting_pos()
-            self.gates.pos, self.gates.quat, self.obstacles.pos = randomize_track(
+        if options.get("check_race_track", True):
+            check_race_track(
                 gates_pos=self.gates.pos,
+                nominal_gates_pos=self.gates.nominal_pos,
                 gates_quat=self.gates.quat,
+                nominal_gates_quat=self.gates.nominal_quat,
                 obstacles_pos=self.obstacles.pos,
+                nominal_obstacles_pos=self.obstacles.nominal_pos,
                 rng_config=self.randomizations,
             )
-            # TODO: The gate safety limits are now hardcoded
-            # 0.5 meters inside the drone safety limits.
-            track_safety_limits = ConfigDict(
-                {"pos_limit_low": self.pos_limit_low[:2], "pos_limit_high": self.pos_limit_high[:2]}
-            )
-            check_gates_in_bound(gates=self.gates, limits=track_safety_limits, tolerance=-0.5)
-
-        elif real_track:
-            self._update_track_poses()
-            if check_track:
-                check_race_track(
-                    gates=self.gates, obstacles=self.obstacles, rng_config=self.randomizations
-                )
-
-        if check_drone:
-            real_pos = self._ros_connector.pos[self.drone_name]
-            nominal_pos = np.array(self.drones.pos[self.rank])
+        if options.get("check_drone_start_pos", True):
             check_drone_start_pos(
-                nominal_pos=nominal_pos,
-                real_pos=real_pos,
+                nominal_pos=self.drones.pos[self.rank],
+                real_pos=self._ros_connector.pos[self.drone_name],
                 rng_config=self.randomizations,
                 drone_name=self.drone_name,
             )
@@ -183,7 +164,7 @@ class RealRaceCoreEnv:
             radio_id=self.rank, radio_channel=self.drone_channel, drone_id=self.drone_id
         )
         self._last_drone_pos_update = 0  # Last time a position was sent to the drone estimator
-        self._reset_env_data(self.data)
+        self.data.reset(np.stack([self._ros_connector.pos[n] for n in self.drone_names]))
         self._reset_drone()
 
         if self.control_mode == "attitude":
@@ -217,7 +198,7 @@ class RealRaceCoreEnv:
         self.data.target_gate += np.asarray(passed)
         self.data.target_gate[self.data.target_gate >= self.n_gates] = -1
         self.data.last_drone_pos[...] = drone_pos
-        self._taken_off |= drone_pos[self.rank, 2] > 0.2
+        self.data.taken_off |= drone_pos[self.rank, 2] > 0.2
         # Send vicon position updates to the drone at a fixed frequency irrespective of the env freq
         # Sending too many updates may deteriorate the performance of the drone, hence the limiter
         if (t := time.perf_counter()) - self._last_drone_pos_update > 1 / self.POS_UPDATE_FREQ:
@@ -327,24 +308,16 @@ class RealRaceCoreEnv:
             for i in range(self.n_obstacles):
                 self.obstacles.pos[i, ...] = pos[f"obstacle{i + 1}"]
         except KeyError as e:
-            raise ValueError(
-                f"Could not find all track objects in the ROS TF tree: {e}. \
-                            Have you enabled the track objects in Vicon/ \
-                            started the motion capture tracking node?"
+            raise KeyError(
+                f"Could not find all track objects in the ROS TF tree: {e}. Have you enabled the "
+                "track objects in Vicon and started the motion capture tracking node?"
             ) from e
-
-    def _update_drone_starting_pos(self):
-        """Update the starting position of the drone from the motion capture system."""
-        drone_pos = self._ros_connector.pos[self.drone_name]
-        drone_quat = self._ros_connector.quat[self.drone_name]
-        drone_pos[2] = 0.01  # Set z to ground level
-        drone_rpy = R.from_quat(drone_quat).as_euler("xyz", degrees=False)
-        self.drones.pos[self.rank], self.drones.rpy[self.rank] = drone_pos, drone_rpy
 
     def _connect_radio(self, radio_id: int, radio_channel: int, drone_id: int):
         """Connect to the drone via radio.
 
-        If the drone is not reachable, a TimeoutError is raised.
+        Raises:
+            TimeoutError: If the drone is not reachable within 10 seconds.
         """
         cflib.crtp.init_drivers()
         uri = f"radio://{radio_id}/{radio_channel}/2M/E7E7E7E7" + f"{drone_id:02x}".upper()
@@ -373,16 +346,8 @@ class RealRaceCoreEnv:
         connected = event.wait(10.0)
         if not connected:
             raise TimeoutError("Timed out while waiting for the drone.")
-        self._connected_to_drone = True
+        self.data.drone_connected = True
         logger.info(f"Drone {drone_id} connected to {uri}")
-
-    def _reset_env_data(self, data: EnvData):
-        """Reset the environment data."""
-        data.target_gate[...] = 0
-        data.gates_visited[...] = False
-        data.obstacles_visited[...] = False
-        drone_pos = np.stack([self._ros_connector.pos[n] for n in self.drone_names])
-        data.last_drone_pos[...] = drone_pos
 
     def _reset_drone(self):
         """Arm the drone, reset estimation."""
@@ -474,31 +439,6 @@ class RealRaceCoreEnv:
                 gate_passed(drone_pos, drone_pos, gate_pos, gate_quat, (0.45, 0.45))
             )
 
-    def update_level_config(self, config: ConfigDict) -> ConfigDict:
-        """Update the level config with the real track objects poses.
-
-        Args:
-            config: A ConfigDict storing the original level configuration.
-
-        Returns:
-            A ConfigDict object, with updated starting pose of gates, obstacles and drones
-        """
-        config = config.copy()
-        # We store the real-world track layout, so randomization must be disabled
-        config.env.track.randomize = False
-        # Overwrite the original positions and orientations with the measured ones
-        for i in range(self.gates.pos.shape[0]):
-            config.env.track.gates.pos[i] = self.gates.pos[i].tolist()
-            config.env.track.gates.rpy[i] = R.from_quat(self.gates.quat[i]).as_euler("xyz").tolist()
-        for i in range(self.obstacles.pos.shape[0]):
-            config.env.track.obstacles.pos[i] = self.obstacles.pos[i].tolist()
-        for i in range(self.drones.pos.shape[0]):
-            config.env.track.drones.pos[i] = self.drones.pos[i].tolist()
-            config.env.track.drones.rpy[i] = (
-                R.from_quat(self.drones.quat[i]).as_euler("xyz").tolist()
-            )
-        return ConfigDict(config)
-
     def close(self):
         """Close the environment.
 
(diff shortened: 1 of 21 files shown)


amacati:
@Rather1337 From a first glance, some files seem to lack ruff formatting. Can we run ruff format in the CI and check if if changed stuff? I think there is a "display changes only" flag that we can use to catch this in PRs.

ratheron:
I says ruff passed in the checks. What do you mean exactly

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -114,52 +120,16 @@ def __init__(
...
-                check_race_track(
-                    self.gates.nominal_pos,
-                    self.gates.nominal_quat,
-                    self.obstacles.nominal_pos,
-                    self.randomizations,
-                )
-        if options.get("check_drone_start_pos", True):
-            pos = self.drones.pos[self.rank, ...]
-            check_drone_start_pos(pos, self.randomizations, self.drone_name)
+        self._prepare_track(deploy_config = options.deploy, track = options.env.track)
```
mschuck-rai: The name "prepare" is pretty undescriptive. I would not be sure what this method does just by looking at the name. Can we make these clearer?

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -265,6 +235,31 @@ def info(self) -> dict:
         """Return an info dictionary containing additional information about the environment."""
         return {}
 
+    def _query_track_obj_pose(self) -> tuple[NDArray, NDArray, NDArray]:
```
mschuck-rai: I am not a huge fan of this method, since it just bundles stuff together that we already should have set up anyways, and its signature is a bit confusing. We say pose, but only return the pos for obstacles? We say obj, but you have to check the signature to know it's gate pos, gate quat, obstacle pos instead of gate pose, obstacle pose, drone pose, etc.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -265,6 +235,31 @@ def info(self) -> dict:
...
+        finally:
+            ros_connector.close()
+        try:
+            for i in range(self.n_gates):
+                gate_pos[i, ...] = pos[f"gate{i + 1}"]
+                gate_quat[i, ...] = quat[f"gate{i + 1}"]
+            for i in range(self.n_obstacles):
+                obstacle_pos[i, ...] = pos[f"obstacle{i + 1}"]
+        except KeyError as e:
+            logger.error(f"Could not find track objects in the ROS TF tree: {e}. Have you enabled the track objects in Vicon/ started the motion capture tracking node?")
```
mschuck-rai: This line blows the 100 chars limit by a fair amount.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
             # TODO: The estimators can't handle state commands, so we simply don't send anything
             # Make sure to use the legacy estimator with the state interface
 
-    def _connect_to_drone(self, radio_id: int, radio_channel: int, drone_id: int) -> Crazyflie:
+    def _prepare_track(self, deploy_config: ConfigDict, track: ConfigDict):
```
mschuck-rai: See comment on naming further up.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
...
+        assert (not track.get("randomize", False)) or (track.get("randomize", False) and deploy_config.get("real_track_objects", True)), "With a randomly generated track, the deploy.real_track_objects must be set to true."
```
mschuck-rai: Checks beyond sanity asserts should be real errors, not asserts. This is not yet consistent across the code, but we should stick to it for new code.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
...
+                )
+
+            # With generated track, the track objects are by nature in the randomization tolerance —— no need to check track!
+            else:
+                if deploy_config.get("check_race_track", True):
+                    check_race_track(
+                        gates=self.gates,
+                        obstacles=self.obstacles,
+                        rng_config=self.randomizations,
+                )
```
mschuck-rai: It should be possible to clean this block up to be tidier. As it is, it's hard to read and reason about
```suggestion
        if not deploy_config.get("real_track_objects"):
            return
        self.gates.pos, self.gates.quat, self.obstacles.pos = self._query_track_obj_pose()
        if track.randomize:
            self.gates.nominal_pos, self.gates.nominal_quat, self.obstacles.nominal_pos = randomize_track(
                gates_pos=self.gates.pos,
                gates_quat=self.gates.quat,
                obstacles_pos=self.obstacles.pos,
                rng_config=self.randomizations
            )
            # TODO: The gate safety limits are now hardcoded: 0.5 meters inside the drone safety limits.
            limit_low = np.array(track.safety_limits.pos_limit_low[:2]) + 0.5
            limit_high = np.array(track.safety_limits.pos_limit_high[:2]) - 0.5
            check_gates_layout(gates=self.gates, limits={'pos_limit_low' : limit_low, 'pos_limit_high' : limit_high})
        if deploy_config.get("check_race_track", True):
                check_race_track(
                    gates=self.gates,
                    obstacles=self.obstacles,
                    rng_config=self.randomizations,
            )
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
...
+    
+    def _prepare_drones(self, deploy_config: ConfigDict, track: ConfigDict):
```
mschuck-rai: Naming comment applies

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
...
+        if track.randomize:
+            # The starting position drones.pos has must be set here even though they are not used in _step(). Since they are the target pose for return controller.
```
mschuck-rai: This is way >100 lines. @Rather1337 can a HiWi double-check how this passes ruff?

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -362,6 +423,22 @@ def _apply_drone_settings(self):
...
         time.sleep(0.1)  # Wait for settings to be applied
 
+    def _query_drone_pose(self) -> tuple[NDArray, NDArray]:
+        ros_connector = ROSConnector(
+            tf_names=[self.drone_name],
+            timeout=5.0,
+        )
+        try:  # Make sure to close the connection if anything goes wrong
+            pos = ros_connector.pos[self.drone_name]
+
```
mschuck-rai:
```suggestion
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -362,6 +423,22 @@ def _apply_drone_settings(self):
         self.drone.param.set_value("flightmode.stabModeYaw", 1)
         time.sleep(0.1)  # Wait for settings to be applied
 
+    def _query_drone_pose(self) -> tuple[NDArray, NDArray]:
```
mschuck-rai: I don't think this is a good idea. We have a connector running anyways that should track the drone. Why are we creating more? They are heavy objects with substantial init times.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
...
+            drone_pos, drone_quat = self._query_drone_pose()
+            drone_pos[2] = 0.01  # Set z to ground level
+            drone_rpy = R.from_quat(drone_quat).as_euler("xyz", degrees=False)
+            self.drones.pos[self.rank], self.drones.rpy[self.rank] = drone_pos, drone_rpy
+
+        else:
+            if deploy_config.get("check_drone_start_pos", True):
+                real_pos, real_quat = self._query_drone_pose()
+                nominal_pos = np.array(self.drones.pos[self.rank])
+                check_drone_start_pos(nominal_pos=nominal_pos, real_pos=real_pos, rng_config=self.randomizations, drone_name=self.drone_name)
```
mschuck-rai: I'm happy to sacrifice a few ms for the sanity check. Also, it's nice to reduce logic complexity.

```suggestion
        if deploy_config.get("check_drone_start_pos", True):
            real_pos, real_quat = self._query_drone_pose()
            nominal_pos = np.array(self.drones.pos[self.rank])
            check_drone_start_pos(nominal_pos=nominal_pos, real_pos=real_pos, rng_config=self.randomizations, drone_name=self.drone_name)
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,7 +284,73 @@ def send_action(self, action: NDArray):
...
+        
+        self._establish_drone_connection(
+            radio_channel=self.drone_ch, drone_id=self.drone_id
+        )
+
+    def _establish_drone_connection(self, radio_channel: int, drone_id: int):
```
mschuck-rai: Please keep namings as simple as possible. Wanna connect to the drone? -> `connect_to_drone`

```suggestion
    def _connect_to_drone(self, radio_channel: int, drone_id: int):
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -114,52 +120,16 @@ def __init__(
         self.data = EnvData.create(
             n_drones=self.n_drones, n_gates=self.n_gates, n_obstacles=self.n_obstacles
         )
-        # Establish drone connection
-        self._drone_healthy = mp.Event()
```
mschuck-rai, line 118: Please do not remove attributes that are set outside of `__init__`. Every self attribute that is set at some point should already be initialized here. Also, why did you remove the `ROSConnector`?

lsy_drone_racing/envs/real_race_env.py
mschuck-rai: It's just a wrapping for 
`pos, quat = self._ros_connector.pos[self.drone_name], self._ros_connector.quat[self.drone_name]` ?

lsy_drone_racing/utils/utils.py
```diff
@@ -79,6 +80,60 @@ def load_config(path: Path) -> ConfigDict:
...
+    
+    """
+    assert output_path.suffix == ".toml", f"Configuration file has to be a TOML file: {output_path}"
+    dict_to_store : dict = {
+            'controller' : config.controller.to_dict(),
+            'deploy' : config.deploy.to_dict(),
+            'env' : config.env.to_dict(),
+            'sim' : config.sim.to_dict()
+    }
+    # Now the randomized generation should be set to false
```
mschuck-rai: Please do not comment what the code does. This is just a repeat of the line below. Comment on **why** you are doing this.

lsy_drone_racing/utils/utils.py
```diff
@@ -79,6 +80,60 @@ def load_config(path: Path) -> ConfigDict:
...
+        {
+            'pos' : env.unwrapped.drones.pos[0].tolist(),
+            'rpy' : env.unwrapped.drones.rpy[0].tolist(),
+            'vel' : [0.0, 0.0, 0.0],
+            'ang_vel' : [0.0, 0.0, 0.0]
+        }
+    ]
+    with open(output_path, "w") as f:
+        toml.dump(dict_to_store, f)
+
```
mschuck-rai: This is _very_ hard to read. Please refactor before merging. Why was this function necessary in the first place?

scripts/deploy.py
```diff
@@ -17,24 +17,27 @@
...
     Args:
         config: Path to the competition configuration. Assumes the file is in `config/`.
         controller: The name of the controller file in `lsy_drone_racing/control/` or None. If None,
             the controller specified in the config file is used.
+        save_track: Path to store the current competition configuration. The file is stored in `config/`. If None, the configuration will not be saved.
     """
     rclpy.init()
     config = load_config(Path(__file__).parents[1] / "config" / config)
+    # Overwrite the config file controller
+    config.controller.file = controller or config.controller.file
```
mschuck-rai: Do not use `or` in case of `None`. Use explicit `is None` checks.

scripts/deploy.py
```diff
@@ -44,17 +47,19 @@ def main(config: str = "level2.toml", controller: str | None = None):
         sensor_range=config.env.sensor_range,
         control_mode=config.env.control_mode,
     )
+
```
mschuck-rai:
```suggestion
```

lsy_drone_racing/utils/checks.py
```diff
@@ -15,63 +14,101 @@
...
+    # TODO: Now the gate check should consider rotation in roll and pitch as well.
+    ang_tol = rng_config.gate_rpy.kwargs.maxval[2]  # Only check yaw rotation
+    for i, (quat, nominal_quat) in enumerate(zip(gates.quat, gates.nominal_quat)):
+        gate_rot = R.from_quat(quat)
+        nominal_rot = R.from_quat(nominal_quat)
+        check_rotation(f"gate{i + 1}", gate_rot, nominal_rot, ang_tol)
+
+    low, high = rng_config.obstacle_pos.kwargs.minval, rng_config.obstacle_pos.kwargs.maxval
+    for i, (pos, nominal_pos) in enumerate(zip(obstacles.pos, obstacles.nominal_pos)):
+        check_bounds(f"obstacle{i + 1}", np.array(pos[:2]), np.array(nominal_pos[:2]), np.array(low[:2]), np.array(high[:2]))
```
mschuck-rai: Do we need all the np casts? They make the code hard to read

mschuck-rai (changes requested):
A couple of comments on the PR. It would be great if you could confirm that the options (with/without track checks etc.) have been tested in real.

mschuck-rai:
> I says ruff passed in the checks. What do you mean exactly

The ruff CI checks with `ruff check`, but `ruff format` would make changes to the files. We should check that PRs have run the latter. E.g. there are instances of two line breaks between class methods etc. We should catch this.

mschuck-rai:
Thanks for working on this btw @N0OBSTUDENT

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: Do you have any suggestion on this? It involves updating the actual gate pose, obstacle position from ROSConnector and checks the track. I can't think of a name that summarize these steps.

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: We need that info to update the pose of the gates and obstacles  and check the track. Originally it was written in check.py and I tried to decouple ROSConnector with check functions.
What about returning a ConfigDict that includes these info? That will make this clearer.

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: The ROSConnector was also created twice in the original code, but the first time was in check.py . I thought that's done intentionally so I copied the procedure here.

In my code, the `self._ros_connector` will not be initialized until gate, obstacle and drone pose checks are passed. So I made a temporary connector here which does not send commands but only query pose of the drone. But I could certainly initialize `self._ros_connector` at the first place.

lsy_drone_racing/envs/real_race_env.py
amacati: Then maybe it should be two steps :D `self._update_track_poses` and `self._check_race_track` seem pretty descriptive

lsy_drone_racing/envs/real_race_env.py
amacati: Do we even need to wrap it in a separate function?

lsy_drone_racing/utils/utils.py
N0OBSTUDENT: To save the current layout of the track.
In level-3  there are "completely randomly generated" tracks and in deployment stage the track is defined by the placement of the gates and obstacles in real-world, i.e. not from the config toml file. So the students may want to rerun the case in simulation but they don't have the track layout. The function could backup the real-world setting for them to rerun in sim.

scripts/deploy.py
N0OBSTUDENT: Will fix that. I thought you thought it was a good practice since it was there in the original one.

lsy_drone_racing/utils/checks.py
N0OBSTUDENT: Hmm no. I did that just to comply to the function signature. I could make argument of `check_bounds()` be `ArrayLike` and cast them in that.

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: OKay, then I do not even need the query track function. It's executed once. I separated everything just to make functions not too long.

lsy_drone_racing/utils/utils.py
N0OBSTUDENT: But I will try to make it look nicer.

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: That's because I could not think of a better way to work around the problem here:

Any race track check failure in `_reset()` will throw an Exception. The Exception will trigger the try-catch-finally in deploy.py, which executes  `return_to_start()` and codes killing the drone. However, the drone is planned be connected at the very last step of `_reset()`, so obviously there is no connection to the drone by that time, which will throw two extremely long exceptions, making the console display a giant stack of Exceptions.

My current work-around is just setting them as `None` upon initialization of the class and initialize them in `_reset()`, so I could have a simple if-else in `self.close()` to judge whether to trigger the return controller and kill the drone.

The other work-around I thought of at the very beginning is that we could do all stuffs except drone connection in `__init__`, since there is certainly no `env.close` called by then. But the problem is that, track checking requires config.deploy field, which is currently not an argument to `__init__` . I suppose you are not willing to add more arguments, so I did not do that.


Edit: Maybe I just need to move the track checking  and drone position checking out of the class, and execute them before initializing the Environment object. Then there will not be a god class there.

Do you have any advice on that?

N0OBSTUDENT:
> A couple of comments on the PR. It would be great if you could confirm that the options (with/without track checks etc.) have been tested in real.

I have tested all cases I could think of  (w or w/o real track objects, w or w/o track check, w or w/o fully randomly generated tracks) in the lab, but I will test them and commit again after reorganizing the codes according to the comments. Could I further add commits upon this PR? Or I can only re-open one with new commits?

amacati:
> Could I further add commits upon this PR? Or I can only re-open one with new commits?

Just keep pushing to your branch, it will get auto-updated here

lsy_drone_racing/envs/real_race_env.py
amacati: I'm not sure why it was created twice, but we should try to reuse them as much as possible without becoming unergonomic.

lsy_drone_racing/envs/real_race_env.py
amacati: I think both is reasonable. It's fine that the environment asserts everything is as expected before starting, but we can also pull that out. Regarding the return to start, it should be feasible to check if the drone has taken off or not, right? At worst, this can be a flag that is set or determined by checking if the drone is close to the ground and moving.

lsy_drone_racing/utils/utils.py
amacati: I like the idea

scripts/deploy.py
amacati: No, it's a bad habit of mine, and I try to get rid off it because it looks so concise, but the outcomes are different. E.g. `1 if y is None else y` with `y = ""` will yield `""`, but `y or 1` will yield `1`.

lsy_drone_racing/utils/utils.py
N0OBSTUDENT: Would it be a good practice to have this in the `RealRaceCoreEnv`? I could write something like `get_level_config()->Dict[str, Any]` in `RealRaceCoreEnv`, which just outputs the dict of updated configuration and we could dump that dict later in deploy.py. Then I do not need  so many arguments and those ugly unwrapping operations.

N0OBSTUDENT:
@amacati Hi, I have modified the code accordingly. Please take a look and see whether it needs any further improvements.
For test cases please refer to [this](https://docs.google.com/spreadsheets/d/1a_WvakzvkbI9FE5ZeVwlMI_acAK3nL8hTaW2EinpsPs/edit?usp=sharing)

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -102,65 +107,40 @@ def __init__(
         self.sensor_range = sensor_range
         self.drone_names = [f"cf{drone['id']}" for drone in drones]
         self.drone_name = self.drone_names[rank]
-        self.channel = drones[rank]["channel"]
+        self.drone_ch = drones[rank]["channel"]
```
amacati: Please leave this as drone channel. I would have to guess what `self.drone_ch` is, but `self.drone_channel` is very clear and not overly verbose.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -102,65 +107,40 @@ def __init__(
...
-        # Establish drone connection
+        self.drone = Crazyflie(rw_cache=str(Path(__file__).parent / ".cache"))
         self._drone_healthy = mp.Event()
-        self._drone_healthy.set()
-        self.drone = self._connect_to_drone(
-            radio_id=rank, radio_channel=drones[rank]["channel"], drone_id=drones[rank]["id"]
-        )
-        self._last_drone_pos_update = 0  # Last time a position was sent to the drone estimator
-
+        self._drone_healthy.clear()
```
amacati: Why is the drone being set to not healthy? I do not see an error state

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -208,11 +188,13 @@ def obs(self) -> dict[str, NDArray]:
...
-        gates_quat = np.where(mask, self.gates.quat, self.gates.nominal_quat).astype(np.float32)
-        mask = self.data.obstacles_visited[..., None]
-        obstacles_pos = np.where(mask, self.obstacles.pos, self.obstacles.nominal_pos).astype(
+        mask_gate = self.data.gates_visited[..., None]
+        gates_pos = np.where(mask_gate, self.gates.pos, self.gates.nominal_pos).astype(np.float32)
+        gates_quat = np.where(mask_gate, self.gates.quat, self.gates.nominal_quat).astype(
+            np.float32
+        )
+        mask_obs = self.data.obstacles_visited[..., None]
+        obstacles_pos = np.where(mask_obs, self.obstacles.pos, self.obstacles.nominal_pos).astype(
```
amacati: Was there a reason for this change? Was it not clear what the code was doing?

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+                    "the deploy.real_track_objects should always be set to True."
+                )
+            self.gates.nominal_pos, self.gates.nominal_quat, self.obstacles.nominal_pos = (
+                randomize_track(
+                    gates_pos=self.gates.pos,
+                    gates_quat=self.gates.quat,
+                    obstacles_pos=self.obstacles.pos,
+                    rng_config=self.randomizations,
+                )
+            )
```
amacati: I am not a fan of the mixing in here and in `_check_track_layout`. This function gets long and does two things depending on the settings. Also, if the function starts with if ...: and skips everything if that condition is not set, it's best to do the check before calling the function and then do the stuff that the function promises.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+                for i in range(self.n_gates):
+                    self.gates.pos[i, ...] = pos[f"gate{i + 1}"]
+                    self.gates.quat[i, ...] = quat[f"gate{i + 1}"]
+                for i in range(self.n_obstacles):
+                    self.obstacles.pos[i, ...] = pos[f"obstacle{i + 1}"]
+            except KeyError as e:
+                logger.error(
+                    f"Could not find track objects in the ROS TF tree: {e}. \
+                             Have you enabled the track objects in Vicon/ \
+                             started the motion capture tracking node?"
```
amacati:
```suggestion
                    f"Could not find track objects in the ROS TF tree: {e}. Have you enabled the "
                    "track objects in Vicon and started the motion capture tracking node?"
```

amacati: It's also not clear why you would want to catch, log, and then reraise the error. Were students unsure what was happening?
In general, try to use as few try-except blocks as possible. If something fails early that's a good thing. There are exceptions of course, e.g. the try finally around the connector to prevent processes hanging indefinitely

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
             # TODO: The estimators can't handle state commands, so we simply don't send anything
             # Make sure to use the legacy estimator with the state interface
 
-    def _connect_to_drone(self, radio_id: int, radio_channel: int, drone_id: int) -> Crazyflie:
+    def _update_track_poses(self, deploy_config: ConfigDict, track: ConfigDict):
+        """Update the track poses from the motion capture system."""
+        if deploy_config.get("real_track_objects", True):
```
amacati: This should be checked before the function is called

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+                {
+                    "pos_limit_low": np.array(track.safety_limits.pos_limit_low[:2]) + 0.5,
+                    "pos_limit_high": np.array(track.safety_limits.pos_limit_high[:2]) - 0.5,
+                }
+            )
+            check_gates_in_bound(gates=self.gates, limits=gate_safety_limits)
+        if deploy_config.get("check_race_track", True):
+            check_race_track(
+                gates=self.gates, obstacles=self.obstacles, rng_config=self.randomizations
+            )
```
amacati: Same thing with the mixing and the checking if we actually do something. The functions themselves make a lot of sense and also make the code in reset much nicer, but this mixing should be sorted out

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+        the function will only overwrite the current drone position with real position.
+        Otherwise, the function will check whether the drone position is within the tolerance limits.
+        """
+        if track.randomize:
+            # The starting position drones.pos has must be set here even though they are not used in _step().
+            # They are the target pose for return controller.
+            drone_pos, drone_quat = (
+                self._ros_connector.pos[self.drone_name],
+                self._ros_connector.quat[self.drone_name],
+            )
```
amacati: Avoid multi-line statements whenever possible. It makes the code so much more readable

```suggestion
            drone_pos = self._ros_connector.pos[self.drone_name]
            drone_quat = self._ros_connector.quat[self.drone_name]
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+            drone_pos[2] = 0.01  # Set z to ground level
+            drone_rpy = R.from_quat(drone_quat).as_euler("xyz", degrees=False)
+            self.drones.pos[self.rank], self.drones.rpy[self.rank] = drone_pos, drone_rpy
+
+        if deploy_config.get("check_drone_start_pos", True):
+            real_pos, _ = (
+                self._ros_connector.pos[self.drone_name],
+                self._ros_connector.quat[self.drone_name],
+            )
```
amacati: ...
```suggestion
            real_pos = self._ros_connector.pos[self.drone_name]
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+        if track.randomize:
+            # The starting position drones.pos has must be set here even though they are not used in _step().
+            # They are the target pose for return controller.
+            drone_pos, drone_quat = (
+                self._ros_connector.pos[self.drone_name],
+                self._ros_connector.quat[self.drone_name],
+            )
+            drone_pos[2] = 0.01  # Set z to ground level
+            drone_rpy = R.from_quat(drone_quat).as_euler("xyz", degrees=False)
+            self.drones.pos[self.rank], self.drones.rpy[self.rank] = drone_pos, drone_rpy
```
amacati: Not sure why this is in here. We are checking if the drone is ready to start. Overwriting the position should not be mixed with the check. Same as for the other functions.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -390,8 +468,6 @@ def wait_for_action(dt: float):
         pos = self._ros_connector.pos[self.drone_name]
         vel = self._ros_connector.vel[self.drone_name]
         # This quick check prevents us from engaging the return controller if we havent even started yet.
```
amacati: This comment does not make sense anymore

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -418,6 +494,43 @@ def _jit(self):
...
+        result_config["env"]["track"]["drones"] = [
+            {
+                "pos": (self.drones.pos[i]).tolist(),
+                "rpy": (R.from_quat(self.drones.quat[i]).as_euler("xyz", degrees=False)).tolist(),
+                "vel": [0.0, 0.0, 0.0],
+                "ang_vel": [0.0, 0.0, 0.0],
+            }
+            for i in range(self.drones.pos.shape[0])
+        ]
+        return ConfigDict(result_config)
```
amacati: Does this not work? Again, we want as less indentation, brackets etc as possible to improve readability.
```suggestion
        config = config.copy()
        # We store the real-world track layout, so randomization must be disabled
        config.env.track.randomize = False
        # Overwrite the original positions and orientations with the measured ones
        for i in range(self.gates.pos.shape[0]):
            config.env.track.gates.pos[i] = self.gates.pos[i].tolist()
            config.env.track.gates.rpy[i] = R.from_quat(self.gates.quat[i]).as_euler("xyz").tolist()
        for i in range(self.obstacles.pos.shape[0]):
            config.env.track.obstacles.pos[i] = self.obstacles.pos[i].tolist()
        for i in range(self.drones.pos.shape[0]):
            config.env.track.drones.pos[i] = self.drones.pos[i].tolist()
            config.env.track.drones.rpy[i] = (
                R.from_quat(self.drones.quat[i]).as_euler("xyz").tolist()
            )
        return ConfigDict(config)
```

lsy_drone_racing/utils/checks.py
```diff
@@ -3,10 +3,9 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
```
amacati: Do not use Unions. Use pipe syntax instead.

lsy_drone_racing/utils/checks.py
```diff
@@ -16,66 +15,125 @@
 logger = logging.getLogger("rosout." + __name__)
 
 
-def check_race_track(
+def randomize_track(
```
amacati: The file is called `checks.py`. This is not a check.

lsy_drone_racing/utils/checks.py
```diff
@@ -16,66 +15,125 @@
...
+
+    obstacle_pos_rand = rng_config.obstacle_pos.kwargs
+    sample_obs_pos = np.random.uniform(
+        obstacle_pos_rand.minval, obstacle_pos_rand.maxval, size=(n_obstacles, 3)
+    )
+    randomized_obstacle_pos = obstacles_pos + sample_obs_pos
+    return randomized_gate_pos, randomized_gate_quat, randomized_obstacle_pos
+
+
+def check_gates_in_bound(gates: ConfigDict, limits: ConfigDict):
```
amacati: Please do not use `ConfigDict`s everywhere. It gets hard to understand what the function does. If you want to check positions, there is no need to pass around dictionaries with nested data. Just pass in the plane data as a `NDArray`.

lsy_drone_racing/utils/checks.py
```diff
@@ -16,66 +15,125 @@
...
+            raise RuntimeError(
+                f"gate{i + 1} position {pos[:2]} is below the predefined minimum limit {limits.pos_limit_low}"
+            )
+        if np.any(pos[:2] > limits.pos_limit_high):
+            raise RuntimeError(
+                f"gate{i + 1} position {pos[:2]} is above the predefined maximum limit {limits.pos_limit_high}"
+            )
+
+
+def check_race_track(gates: ConfigDict, obstacles: ConfigDict, rng_config: ConfigDict):
```
amacati: Same here

lsy_drone_racing/utils/checks.py
```diff
@@ -16,66 +15,125 @@
...
+
+    Raise error when the actual value exceeds the low or high boundaries.
+
+    Args:
+        name: Name of the object being checked.
+        actual: NDArray of real value and is expected to be [N, ].
+        desired: NDArray of nominal value and is expected to be [N, ].
+        low: NDArray of the minimum value of (actual - desired) and is expected to be [N, ]
+        high: NDArray of the maximum value of (actual - desired) and is expected to be [N, ]
+    """
```
amacati, line 90: Nothing in this function assumes a shape. Also, do not repeat types in the doc string. They are contained in the function signature already.

```suggestion
    """Check if the actual value is within the specified bounds of the desired value.

    Args:
        name: Name of the object being checked.
        actual: Values to check.
        desired: Reference values.
        low: Lower bound. Minimum permissible value of (actual - desired).
        high: Upper bound. Maximum value of (actual - desired).
    
    Raises:
            RuntimeError: The values are not in the permissible interval.
    """
```

lsy_drone_racing/utils/checks.py
```diff
@@ -86,11 +144,46 @@ def check_bounds(name: str, actual: NDArray, desired: NDArray, low: NDArray, hig
...
-        raise RuntimeError(
-            f"{name} exceeds rotation tolerances ({ang_tol}).\n"
-            f"Rotation is: {actual}, should be: {desired}"
-        )
+def check_rotation(
+    name: str,
+    actual_rot: R,
+    desired_rot: R,
+    low: Union[np.ndarray, float],
+    high: Union[np.ndarray, float],
```
amacati:
```suggestion
    low: NDArray | float,
    high: NDArray | float,
```

lsy_drone_racing/utils/checks.py
```diff
@@ -86,11 +144,46 @@ def check_bounds(name: str, actual: NDArray, desired: NDArray, low: NDArray, hig
...
+        if np.any(diff < low):
+            raise RuntimeError(
+                f"{name} exceeds lower rotation tolerances ({low}).\n"
+                f"Rotation is: {actual}, should be: {desired}"
+            )
+        elif np.any(diff > high):
+            raise RuntimeError(
+                f"{name} exceeds higher rotation tolerances ({high}).\n"
+                f"Rotation is: {actual}, should be: {desired}"
+            )
```
amacati: This needs to be revised. We cannot change the metric of angular measurement and with it the whole behaviour of the function depending on whether you pass in a 3D array or a float. This is error-prone and confusing at best.

scripts/deploy.py
```diff
@@ -25,16 +26,20 @@
...
 
-def main(config: str = "level2.toml", controller: str | None = None):
+def main(config: str = "level2.toml", controller: str | None = None, save_track: str | None = None):
     """Deployment script to run the controller on the real drone.
 
     Args:
         config: Path to the competition configuration. Assumes the file is in `config/`.
         controller: The name of the controller file in `lsy_drone_racing/control/` or None. If None,
             the controller specified in the config file is used.
+        save_track: Path to store the current competition configuration. The file is stored in `config/`. If None, the configuration will not be saved.
```
amacati: `save_track` sounds like a flag, not a path.

scripts/deploy.py
```diff
@@ -25,16 +26,20 @@
...
     """
     rclpy.init()
     config = load_config(Path(__file__).parents[1] / "config" / config)
+    if controller is not None:
+        config.controller.file = controller
```
amacati, line 39: config.controller.file should be a `Path` if possible

scripts/deploy.py
```diff
@@ -44,17 +49,21 @@ def main(config: str = "level2.toml", controller: str | None = None):
...
     )
+
     try:
-        options = {
-            "check_drone_start_pos": config.deploy.check_drone_start_pos,
-            "check_race_track": config.deploy.check_race_track,
-            "real_track_objects": config.deploy.real_track_objects,
-        }
-        obs, info = env.reset(options=options)
+        obs, info = env.reset(options=config)
```
amacati: I am not a fan of passing the complete config as options. We do not want god objects that contain everything, and if possible we should also limit the use of god config objects. Why was this changed?

scripts/deploy.py
```diff
@@ -44,17 +49,21 @@ def main(config: str = "level2.toml", controller: str | None = None):
...
+        if save_track is not None:
+            output_path = Path(__file__).parents[1] / "config" / save_track
+            assert output_path.suffix == ".toml", (
+                f"Configuration file has to be a TOML file: {output_path}"
+            )
```
amacati:
```suggestion
            if not output_path.suffix == ".toml":
                raise ValueError(f"Configuration file has to be a TOML file: {output_path}")
```

amacati: Prefer real errors over asserts if the error is a likely possibility. Asserts should not be part of the intended control flow. They are meant to sanity check stuff. In fact, you can turn off all asserts with `python -O ...`.

amacati (commented):
Looks much better, but there are multiple things that still need to be addressed.

amacati:
Please also resolve any comments I made that you fixed or addressed to keep the conversation clean.

The real-world test sheet looks amazing. Thanks a lot for going through all of the options.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -289,36 +271,130 @@ def send_action(self, action: NDArray):
...
+    def _check_track_layout(self, deploy_config: ConfigDict, track: ConfigDict):
+        """Check the track layout for validity.
+
+        Note:
+            If the randomly generation mode is set to true,
+            the function will check whether the gates are within the safety limits.
+            Otherwise, the function will check whether the track is within the tolerance limits.
+        """
+        if track.randomize:
+            # TODO: The gate safety limits are now hardcoded: 0.5 meters inside the drone safety limits.
```
amacati: This should not be hard coded :D

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: Okay, I suppose it was a existing practice

lsy_drone_racing/utils/checks.py
N0OBSTUDENT: But here if I pass the nominal gate positions, gate rotations, obstacle positions, randomization upper/lower bound of gate positions, rotations, obstacle positions, that makes the signature very long. What should be `ConfigDict` and what should not?

lsy_drone_racing/utils/checks.py
N0OBSTUDENT: That's due to legacy concerns. Shall I just keep the 3D array? The scalar is actually not used now

scripts/deploy.py
N0OBSTUDENT: Now to check whether the track is in place, both the field `track.randomize` and `deploy` are needed. This is because the `track.randomize` field gives out whether the track is randomly generated, and this is necessary for checking procedures.

The most important thing is that I need to check whether the `deploy.real_track_object` is `True` when `track.randomize` is set to `True`. If that's skippable, I can do everything regarding randomly generated track in `__init__` and get rid of `deploy` field.

I can not think of any work around, except changing the toml file structure.

scripts/deploy.py
N0OBSTUDENT: The `config.controller.file` is originally a string loaded from the toml file
``` 
[controller]
file = "state_controller.py"
```
And I do not think that the type should be changed. It has to be `str` in order to get concatenated afterwards
```  
control_path = Path(__file__).parents[1] / "lsy_drone_racing/control"
controller_path = control_path / config.controller.file
```

amacati (approved):

PR #58, opened 2025-12-12
Fix action range definition in race_core for attitude interface.

yufei4hua:
Aligned the build action space function with the one in the default drone_env.
Added drone_model: str as input, then read thrust_min & max from ForceTorqueParams.


amacati (approved):
Looks good, thanks!

Issue #59, opened 2025-12-13
Inconsistent thrust bounds between ForceTorqueParams.load() and load_params() for the same drone model


Yuming-Lee24:
Hi dear Teaching Team:
I have encountered an issue:
## Description

When loading thrust parameters for the same drone model (`cf21B_500`), two different loading methods return different values:

| Source | `thrust_min * 4` | `thrust_max * 4` |
|--------|------------------|------------------|
| `VecDroneRaceEnv` / `ForceTorqueParams.load()` | 0.12 | 0.72 |
| `load_params()` from `drone_models.core` | 0.0855 | 0.80 |

This causes a mismatch between the action space bounds used by the environment and those used by controllers (e.g., `attitude_rl.py`), leading to incorrect action scaling during inference.

## Impact

- RL policies trained with normalized actions `[-1, 1]` will be incorrectly mapped to physical thrust values
- Controllers using `load_params()` will assume different thrust limits than the environment enforces
- This can cause unexpected flight behavior or suboptimal performance

## Reproduction
```python
from pathlib import Path
from lsy_drone_racing.utils import load_config
from lsy_drone_racing.envs.drone_race import VecDroneRaceEnv
from drone_models.core import load_params
from drone_controllers.mellinger.params import ForceTorqueParams

config = load_config(Path("config/level0.toml"))

# Method 1: Environment (uses ForceTorqueParams internally)
env = VecDroneRaceEnv(num_envs=1, freq=config.env.freq, sim_config=config.sim,
                      track=config.env.track, sensor_range=config.env.sensor_range,
                      control_mode="attitude", seed=42, device="cpu")
print(f"Env bounds: [{env.single_action_space.low[3]}, {env.single_action_space.high[3]}]")
# Output: [0.12, 0.72]

# Method 2: load_params (used in attitude_rl.py)
drone_params = load_params(config.sim.physics, config.sim.drone_model)
print(f"load_params bounds: [{drone_params['thrust_min']*4}, {drone_params['thrust_max']*4}]")
# Output: [0.0855, 0.8]

# Method 3: Direct ForceTorqueParams
params = ForceTorqueParams.load("cf21B_500")
print(f"ForceTorqueParams bounds: [{params.thrust_min*4}, {params.thrust_max*4}]")
# Output: [0.12, 0.72]
```
Which loading method should I use? Thanks in advance!

Yuming-Lee24:
[...]

PR #60, opened 2025-12-16
Fix Issue #59: Load thrust limits from drone-controller.

yufei4hua:
Fix tiny bug in deployment script for RL. Load thrust limits params from drone-controller instead of drone-models to keep consistent with action_space() in crazyflow.envs.drone_env, where it also load params from controller.

This is coming from the misalignment of thrust limits in drone-controller and drone-models, which are already fixed in their latest main branches. That means if we simply upgrade to new published packages, these values will also be aligned.

If you think this is not necessary, please discard this PR.

### DIFF
--- a/lsy_drone_racing/control/attitude_rl.py
+++ b/lsy_drone_racing/control/attitude_rl.py
@@ -15,6 +15,7 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 import torch
+from drone_controllers.mellinger.params import ForceTorqueParams
 from drone_models.core import load_params
 from scipy.interpolate import CubicSpline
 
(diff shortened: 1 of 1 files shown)


amacati:
Can we instead update the drone-models / drone-controller packages, or is that not possible because the students are working on a fixed version of the sim that requires the old versions?

yufei4hua:
I think that's the plan! I will open another PR for updating pyproject.toml after crazyflow is ready, where we would also switch to ros2-kilted to solve this numpy==1 issue at the mean time. I've already tested the new environment.
This only affects 1 group that is doing RL. I already told them how to fix this for themselves.

PR #61, opened 2025-12-19
Update for ADR ws25.

yufei4hua:
1. Support level 3 deployment & refactor code.
2. Upgrade to latest drone-models, drone-controllers, drone-estimators and ros2-kilted.
3. Update docs.


ratheron (approved):

mschuck-rai:
Just a small comment @yufei4hua : we don't need to change the drone IDs in the repository. They will change from time to time, and students will have to adapt them anyways.

PR #62, opened 2025-12-19
Fix cam config and add pngs

ratheron:
Fixed:

- Since we bumped the crazyflow version, the cam config changed
- Gate pngs were not properly installed 

@amacati the first one should be checked by running the sim with -r in some test.


amacati (approved):

amacati:
Can you open an issue so we don't forget about the rendering test?

PR #65, opened 2025-12-21
Fixed unexpected UnboundError when trying to close a not initialized ROSConnector

N0OBSTUDENT:
And set SCIPY_ARRAY_API environment variable before importing scipy packages

### DIFF
--- a/lsy_drone_racing/envs/real_race_env.py
+++ b/lsy_drone_racing/envs/real_race_env.py
@@ -297,11 +297,13 @@ class RealRaceCoreEnv:
         """Update the track poses from the motion capture system."""
         tf_names = [f"gate{i}" for i in range(1, self.n_gates + 1)]
         tf_names += [f"obstacle{i}" for i in range(1, self.n_obstacles + 1)]
+        ros_connector = None
         try:
             ros_connector = ROSConnector(tf_names=tf_names, timeout=10.0)
             pos, quat = ros_connector.pos, ros_connector.quat
         finally:
-            ros_connector.close()
+            if ros_connector is not None:
+                ros_connector.close()
         try:
             for i in range(self.n_gates):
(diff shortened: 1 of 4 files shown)


amacati:
Thanks for fixing these. Let me know when you have pushed all your changes and things are ready for review.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -297,11 +297,13 @@ def _update_track_poses(self):
...
         tf_names = [f"gate{i}" for i in range(1, self.n_gates + 1)]
         tf_names += [f"obstacle{i}" for i in range(1, self.n_obstacles + 1)]
+        ros_connector = None
         try:
             ros_connector = ROSConnector(tf_names=tf_names, timeout=10.0)
             pos, quat = ros_connector.pos, ros_connector.quat
         finally:
-            ros_connector.close()
+            if ros_connector is not None:
+                ros_connector.close()
```
ratheron, line 306: Fixing the ROSConnector timeout like this, the user wont get any meaningful information about the error directly (only when scrolling up a bit). The program will error out in the next try catch block, since pos and quat is not set. The ROSConnector is already raising meaningful error messages. We should simply forward those

amacati: As I understand it, the creation of `ros_connector` itself can fail in line 301. We then execute the `finally` block. Since `ros_connector` is not set, calling `close` on it will fail. With this fix, we immediately throw instead of confusing users with a `NoneType` has no function `close`. The timeout error from the failed `ros_connector` should still throw after the `finally` block.

N0OBSTUDENT: So as I understand, that's a 👍 for this fix?
There's no other things to be done in this PR. Please consider merging it if it looks good to you.

amacati (approved):
Looks good, thanks!

PR #66, opened 2026-01-06
Fix bug: Takeoff pos randomization is overridden by complete track randomization.

yufei4hua:
1. In update_sim_data, write new takeoff pos to default_data, instead of directly override data.state.pos

2. call sim.reset() after update_sim_data(), to add additional domain randomization. 
- Since randomization on drone pos is by default built in sim.reset_pipeline, I think it's better to keep this structure: randomize drone pos by calling sim.reset()
- Neither the complete randomization nor the small perturbation are based on reset SimData, so it's fine to reset SimData afterwards.
- Also compatible with level 2, where the default position stays the same, then the domain randomization will also take effect.
- At last, we update EnvData, where we need the current drone pos to update target gate and so on.

3. Explicitly write mask as an argument in update_sim_data(), although it's closure captured, put it in args for better readability.

4. Apply mask when overriding data.state.pos. Otherwise, when any world is marked for reset, all worlds will be reset to the newly generated takeoff pos! 
(Should not have this problem if we only write to default data, but still doesn't make sense to update default data when some worlds are not yet done.)

### DIFF
--- a/tests/integration/test_envs.py
+++ b/tests/integration/test_envs.py
@@ -2,10 +2,12 @@ from pathlib import Path
 
 import gymnasium
 import jax
+import jax.numpy as jp
 import pytest
 from drone_models import available_models
 
 import lsy_drone_racing  # noqa: F401, environment registrations
+from lsy_drone_racing.envs.drone_race import DroneRaceEnv
 from lsy_drone_racing.utils import load_config
 
 CONFIG_FILES = {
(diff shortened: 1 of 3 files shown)


amacati (changes requested):
The code changes look good. Can you also add a test that check if this is working as expected? It should fail on the old version and succeed with your fixes.

yufei4hua:
I fixed the randomization of race objects in the multi-world case.
But it might take some time for me to write test cases for them.

amacati:
That's fine, we'll wait for the tests before we merge. Once you have included them, could you please verify locally that the tests fail without your patch?

yufei4hua:
Sure! I will do that. That would be the main point of the tests.

yufei4hua:
Hi! I've written a test for this fix.
I've test it on my machine. It passes for my fix but not for the old versions.
What I did:
1. create 2 worlds (don't care about physics/devices)
2. Turn on autoreset
3. Backup default data (gates, obstacles, drone)
4. env.reset()
5. Check if track is randomized upon first reset
6. env.step() and log track data
7. Manually set reset flag of 2nd word
8. env.step() to trigger autoreset and log track data
9. Check if track is re-randomized for the 2nd world but not for the 1st world
10. Check drone pos should be again perturbed upon default_data, even lv3 should be this case

Please check if the structure is fine, if I need to shorten the error messages or if the test name is reasonable.
Thanks!

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+        "freq": config.env.freq,
+        "sim_config": config.sim,
+        "sensor_range": config.env.sensor_range,
+        "track": config.env.track,
+        "disturbances": config.env.get("disturbances"),
+        "randomizations": config.env.get("randomizations"),
+        "seed": config.env.seed,
+    }
+
+    env: DroneRaceEnv = gymnasium.make_vec("DroneRacing-v0", **kwargs)
```
amacati:
```suggestion
    env: DroneRaceEnv = gymnasium.make_vec(
        "DroneRacing-v0",
        num_envs=2,
        control_mode="attitude",
        freq=config.env.freq,
        sim_config=config.sim,
        track=config.env.track,
        randomizations=config.env.get("randomizations"
    )
```

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+    env.autoreset = True
```
amacati: Should be on by default for vector envs.

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+
+    def get_obj_mocap(env: DroneRaceEnv) -> jax.Array:
+        gate_mocap_ids = env.data.gate_mj_ids
+        obstacle_mocap_ids = env.data.obstacle_mj_ids
+        gates_mocap_pos = env.sim.mjx_data.mocap_pos[:, gate_mocap_ids].copy()
+        gates_mocap_quat = env.sim.mjx_data.mocap_quat[:, gate_mocap_ids][..., [1, 2, 3, 0]].copy()
+        obstacles_mocap_pos = env.sim.mjx_data.mocap_pos[:, obstacle_mocap_ids].copy()
```
amacati: These are immutable by default, should work without copy

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+    )
+    assert not jp.allclose(gates_quat, nominal_gates_quat), (
+        "Track should be randomized, but gates_quat remains default."
+    )
+    assert not jp.allclose(obstacles_pos, nominal_obstacles_pos), (
+        "Track should be randomized, but obstacles_pos remains default."
+    )
+    assert not jp.allclose(drone_pos, nominal_drone_pos), (
+        "Drone takeoff pos should be randomized, but it remains default."
+    )
```
amacati: Would be nice to have them on a single line to make the code more readable. E.g. use "gates_pos wasn't randomized" which is the same message, but more concise

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+
+    action = jp.tile(jp.array([0.0, 0.0, 0.0, 0.6]), (2, 1, 1))  # ensure drone doesn't crash
+    env.step(action)
+    gates_pos_1, gates_quat_1, obstacles_pos_1 = get_obj_mocap(env)
+
+    # Force reset 2nd world
+    env.data = env.data.replace(marked_for_reset=jp.array([0, 1]))
```
amacati: Does this change the array type from bool to int? That would force recompiles and makes the test slower. Otherwise use 


```suggestion
    env.data = env.data.replace(marked_for_reset=env.data.marked_for_reset.at[1].set(1))
```

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+
+    # Step once to trigger autoreset
+    env.step(action)
+    gates_pos_2, gates_quat_2, obstacles_pos_2 = get_obj_mocap(env)
+    drone_pos = env.sim.data.states.pos[:, 0, :].copy()
+    nominal_drone_pos = env.sim.default_data.states.pos[:, 0, :].copy()
```
amacati: These should also be immutable and not require a copy.

tests/integration/test_envs.py
```diff
@@ -114,3 +116,103 @@ def test_multi_drone_envs(config_file: str, physics: str, device: str):
...
+
+    assert not jp.allclose(gates_pos_2[1], gates_pos_1[1]), (
+        "World 1 should be re-randomized, but gates_pos remain unchanged."
+    )
+    assert not jp.allclose(gates_quat_2[1], gates_quat_1[1]), (
+        "World 1 should be re-randomized, but gates_quat remain unchanged."
+    )
+    assert not jp.allclose(obstacles_pos_2[1], obstacles_pos_1[1]), (
+        "World 1 should be re-randomized, but obstacles_pos remain unchanged."
+    )
```
amacati: Would be nice to have all of these on one line each if possible. Makes the test much more readable. E.g. use "Unexpected reset of gates_pos in world 0"

amacati (changes requested):

amacati:
The test looks good, thanks! I had some minor comments on the code, but these should be quick to fix.

tests/integration/test_envs.py
yufei4hua: "set(1)" would raise a warning: "cannot safely cast value from dtype=int32 to dtype=bool with jax_numpy_dtype_promotion=standard. In future JAX releases this will result in an error."

I'll put "env.data.marked_for_reset.at[1].set(True)" also for better readability.

tests/integration/test_envs.py
yufei4hua: I've tried my best, but I don't think it's possible unless I change "obstacles" to "obst" for all variable naming.

amacati: That's fine. If possible, make things brief, but never at the cost of making it more obscure. You can resolve any conversation that you have addressed and lmk when you are done with everything.

yufei4hua: Hi, please review the current version. I think it's already concise and not too complex.
Test function still works.

amacati (approved):

amacati:
Nice, thanks Yufei!

PR #69, opened 2026-01-23
Add action clipping for race envs.

yufei4hua:
Clip action according to self.action_space
Use inf for state interface.

### DIFF
--- a/lsy_drone_racing/envs/drone_race.py
+++ b/lsy_drone_racing/envs/drone_race.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Literal
 
 import gymnasium
-import jax.numpy as jp
 from gymnasium import Env
 from gymnasium.vector import VectorEnv
 from gymnasium.vector.utils import batch_space
@@ -95,7 +94,9 @@ class DroneRaceEnv(RaceCoreEnv, Env):
         Returns:
             Observation, reward, terminated, truncated, and info.
         """
-        action = jp.clip(action, self.action_space.low, self.action_space.high)
+        action = self._sanitize_action(
+            action, self.action_space.low, self.action_space.high, self.num_envs, self.device
+        )
         obs, reward, terminated, truncated, info = self._step(action)
         obs = {k: v[0, 0] for k, v in obs.items()}
         info = {k: v[0, 0] for k, v in info.items()}
@@ -181,7 +182,9 @@ class VecDroneRaceEnv(RaceCoreEnv, VectorEnv):
         Returns:
             Observation, reward, terminated, truncated, and info.
         """
-        action = jp.clip(action, self.action_space.low, self.action_space.high)
+        action = self._sanitize_action(
+            action, self.action_space.low, self.action_space.high, self.num_envs, self.device
+        )
         obs, reward, terminated, truncated, info = self._step(action)
         obs = {k: v[:, 0] for k, v in obs.items()}
         info = {k: v[:, 0] for k, v in info.items()}
--- a/lsy_drone_racing/envs/drone_race.py
+++ b/lsy_drone_racing/envs/drone_race.py
@@ -95,7 +95,7 @@ class DroneRaceEnv(RaceCoreEnv, Env):
             Observation, reward, terminated, truncated, and info.
         """
         action = self._sanitize_action(
-            action, self.action_space.low, self.action_space.high, self.num_envs, self.device
+            action, self.action_space.low, self.action_space.high, 1, 1, self.device
         )
         obs, reward, terminated, truncated, info = self._step(action)
         obs = {k: v[0, 0] for k, v in obs.items()}
@@ -183,7 +183,7 @@ class VecDroneRaceEnv(RaceCoreEnv, VectorEnv):
             Observation, reward, terminated, truncated, and info.
         """
         action = self._sanitize_action(
-            action, self.action_space.low, self.action_space.high, self.num_envs, self.device
+            action, self.action_space.low, self.action_space.high, self.num_envs, 1, self.device
         )
         obs, reward, terminated, truncated, info = self._step(action)
         obs = {k: v[:, 0] for k, v in obs.items()}
--- a/lsy_drone_racing/envs/drone_race.py
+++ b/lsy_drone_racing/envs/drone_race.py
@@ -94,6 +94,7 @@ class DroneRaceEnv(RaceCoreEnv, Env):
         Returns:
             Observation, reward, terminated, truncated, and info.
         """
+        assert action.shape == self.action_space.shape, "Action shape mismatch."
         action = self._sanitize_action(
             action, self.action_space.low, self.action_space.high, 1, 1, self.device
         )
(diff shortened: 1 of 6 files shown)


lsy_drone_racing/envs/drone_race.py
```diff
@@ -94,6 +95,7 @@ def step(self, action: Array) -> tuple[dict, float, bool, bool, dict]:
         Returns:
             Observation, reward, terminated, truncated, and info.
         """
+        action = jp.clip(action, self.action_space.low, self.action_space.high)
```
amacati: Can we move this into the compiled step function? Otherwise it's going to be pretty slow

yufei4hua: Yes, I should better do that.
It actually triggered a bug in the previous test I wrote. I want to know, do we want to check action shape for single/vector envs? Or we just do a reshape in racecore apply_action()?

amacati: I think we should raise an error. If your shapes are off, chances are your NN or MPC or whatever are not doing what you think they are

lsy_drone_racing/envs/race_core.py
```diff
@@ -440,6 +440,14 @@ def apply_action(self, action: Array):
             case _:
                 raise ValueError(f"Unsupported control mode: {self.sim.control}")
 
+    @staticmethod
+    @partial(jax.jit, static_argnames=["n_worlds", "device"])
+    def _sanitize_action(
+        action: Array, low: NDArray, high: NDArray, n_worlds: int, device: str
+    ) -> Array:
+        action = jp.clip(action, low.reshape(action.shape), high.reshape(action.shape))
+        return jp.array(action, device=device).reshape((n_worlds, 1, -1))
```
amacati: This should not be correct. Race core can also have multiple drones

lsy_drone_racing/envs/drone_race.py
yufei4hua: I added a shape assertion before clipping actions.

lsy_drone_racing/envs/race_core.py
yufei4hua: True, I've added n_drones to args.

mschuck-rai: Btw, didn't we say we want to raise instead of trying to fix things?

yufei4hua: You're right, I don't need to reshape in the sanitize fn after asserting the shape. 
Btw, it seems that the XLA's alignment warning is fixed, I'll remove the copy and use 
```
    def apply_action(self, action: Array):
        """Apply the commanded state action to the simulation."""
        # Convert to a buffer that meets XLA's alginment restrictions to prevent warnings. See
        # https://github.com/jax-ml/jax/discussions/6055
        # Tracking issue:
        # https://github.com/jax-ml/jax/issues/29810
        # Forcing a copy here is less efficient, but avoids the warning.
        # action = np.reshape(action, (self.sim.n_worlds, self.sim.n_drones, -1), copy=True)
        action = action.reshape((self.sim.n_worlds, self.sim.n_drones, -1))
```

lsy_drone_racing/envs/drone_race.py
```diff
@@ -94,6 +94,10 @@ def step(self, action: Array) -> tuple[dict, float, bool, bool, dict]:
         Returns:
             Observation, reward, terminated, truncated, and info.
         """
+        assert action.shape == self.action_space.shape, "Action shape mismatch."
+        action = self._sanitize_action(
+            action, self.action_space.low, self.action_space.high, self.device
+        )
```
amacati, line 100: Since we are doing this in every env, we should do it in race_core's _step instead
```suggestion
```

lsy_drone_racing/envs/drone_race.py
```diff
@@ -179,6 +183,10 @@ def step(self, action: Array) -> tuple[dict, Array, Array, Array, dict]:
...
```
amacati, line 189:
```suggestion
```

lsy_drone_racing/envs/multi_drone_race.py
```diff
@@ -103,6 +104,10 @@ def step(self, action: Array) -> tuple[dict, Array, Array, Array, dict]:
         Returns:
             Observation, reward, terminated, truncated, and info for all drones.
         """
+        assert action.shape == self.action_space.shape, "Action shape mismatch."
+        action = self._sanitize_action(
+            action, self.action_space.low, self.action_space.high, self.device
+        )
```
amacati, line 110:
```suggestion
```

lsy_drone_racing/envs/multi_drone_race.py
```diff
@@ -191,4 +197,8 @@ def step(self, action: Array) -> tuple[dict, Array, Array, Array, dict]:
         Args:
             action: Action for all drones, i.e., a batch of (n_drones, action_dim) arrays.
         """
+        assert action.shape == self.action_space.shape, "Action shape mismatch."
+        action = self._sanitize_action(
+            action, self.action_space.low, self.action_space.high, self.device
+        )
```
amacati, line 203:
```suggestion
```

lsy_drone_racing/envs/race_core.py
```diff
@@ -422,12 +422,7 @@ def _step(self, action: Array) -> tuple[dict[str, Array], float, bool, bool, dic
 
     def apply_action(self, action: Array):
         """Apply the commanded state action to the simulation."""
-        # Convert to a buffer that meets XLA's alginment restrictions to prevent warnings. See
-        # https://github.com/jax-ml/jax/discussions/6055
-        # Tracking issue:
-        # https://github.com/jax-ml/jax/issues/29810
-        # Forcing a copy here is less efficient, but avoids the warning.
-        action = np.reshape(action, (self.sim.n_worlds, self.sim.n_drones, -1), copy=True)
+        action = action.reshape((self.sim.n_worlds, self.sim.n_drones, -1))
```
amacati, line 425: If we move the assertion and clipping here, it will automatically apply to all envs
```suggestion
        assert action.shape == self.action_space.shape, "Action shape mismatch."
        action = self._sanitize_action(
            action, self.action_space.low, self.action_space.high, self.device
        )
```

lsy_drone_racing/envs/multi_drone_race.py
```diff
@@ -164,6 +169,7 @@ def __init__(
             device=device,
         )
         self.num_envs = num_envs
+        self.n_drones = n_drones
```
amacati, line 172: Same here

lsy_drone_racing/envs/multi_drone_race.py
```diff
@@ -71,6 +71,7 @@ def __init__(
             max_episode_steps=max_episode_steps,
             device=device,
         )
+        self.n_drones = n_drones
```
amacati, line 74: Do we need to store this? It should be available as self.sim.n_drones, right?

amacati (changes requested):
Sorry, completely forgot about this one. My main point: can't we move the asserts into race core?

lsy_drone_racing/envs/multi_drone_race.py
yufei4hua: Seems that I didn't use it at the end. I'll remove them.

lsy_drone_racing/envs/race_core.py
yufei4hua: Hi, sorry I didn't realize that we can actually read self.action_space defined by subclasses. 
Since we won't use a bare core env, this should work without problem.
I'll move assert in race_core. Then I'll still reshape to the required shape for the core sim interface.

yufei4hua: Please review changes in the new PR.

amacati:
Superseded  by #70.

PR #70, opened 2026-02-06
Implementing takeoff platform for proper crash detecting in simulation.

yufei4hua:
Disable ground contact, use z_pos to check crashing. 
Then define a small region(4cm*4cm) around the takeoff position where drones can stay without crashing.
Modification of the code:
1. disable ground contact check: _load_contact_masks()
2. use pos_limit_low to detect crashing: set data.pos_limit_low.z to 0.0
3. compute not_in_platform, disabled &= not_in_platform, then compute other contacts (with objects)
4. to work with lv2 with drone_pos randomization, added takeoff_pos to EnvData, log takeoff_pos upon reset.
Tested on lv0~3, also with autoreset=True.

I modified gate collisions in this branch as well, cause it's also quite small and should not affect any other behaviors.

### DIFF
--- a/lsy_drone_racing/envs/race_core.py
+++ b/lsy_drone_racing/envs/race_core.py
@@ -282,7 +282,7 @@ class RaceCoreEnv:
             device=device,
         )
         use_box_collision(self.sim, True)
-        self.default_cam_config = {
+        self.cam_config = {
             "distance": sim_config.camera_view[0],
             "azimuth": sim_config.camera_view[1],
             "elevation": sim_config.camera_view[2],
@@ -438,7 +438,7 @@ class RaceCoreEnv:
 
     def render(self):
         """Render the environment."""
-        self.sim.render(default_cam_config=self.default_cam_config)
+        self.sim.render(cam_config=self.cam_config)
 
     def close(self):
         """Close the environment by stopping the drone and landing back at the starting position."""
--- a/lsy_drone_racing/envs/race_core.py
+++ b/lsy_drone_racing/envs/race_core.py
@@ -326,7 +326,7 @@ class RaceCoreEnv:
             obstacle_mj_ids=obstacle_ids,
             max_episode_steps=max_episode_steps,
             sensor_range=sensor_range,
-            pos_limit_low=[-3, -3, -1e-3],
+            pos_limit_low=[-3, -3, 0.0],
             pos_limit_high=[3, 3, 2.5],
             device=self.device,
         )
@@ -413,7 +413,7 @@ class RaceCoreEnv:
         marked_for_reset = self.data.marked_for_reset
         # Apply the environment logic with updated simulation data.
         self.data = self._step_env(
-            self.data, drone_pos, mocap_pos, mocap_quat, contacts, self.sim.freq
+            self.data, drone_pos, mocap_pos, mocap_quat, contacts, self.sim.default_data.states.pos
         )
         # Auto-reset envs. Add configuration option to disable for single-world envs
         if self.autoreset and marked_for_reset.any():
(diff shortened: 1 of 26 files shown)


lsy_drone_racing/envs/race_core.py
```diff
@@ -622,6 +618,9 @@ def _truncated(steps: Array, max_episode_steps: Array, n_drones: int) -> Array:
     def _disabled_drones(pos: Array, contacts: Array, data: EnvData) -> Array:
         disabled = data.disabled_drones | jp.any(pos < data.pos_limit_low, axis=-1)
         disabled = disabled | jp.any(pos > data.pos_limit_high, axis=-1)
+        not_in_platform = jp.any(pos[..., :2] < data.takeoff_pos[..., :2] - 0.02, axis=-1)
+        not_in_platform |= jp.any(pos[..., :2] > data.takeoff_pos[..., :2] + 0.02, axis=-1)
+        disabled = disabled & not_in_platform
```
amacati: This is problematic. With the current code, drones that fly straight up into the sky won't get terminated.

```suggestion
        not_in_platform = jp.any(pos[..., :2] < data.takeoff_pos[..., :2] - 0.02, axis=-1)
        not_in_platform |= jp.any(pos[..., :2] > data.takeoff_pos[..., :2] + 0.02, axis=-1)
        disabled = data.disabled_drones | jp.any(pos < data.pos_limit_low, axis=-1) & not_in_platform
        disabled = disabled | jp.any(pos > data.pos_limit_high, axis=-1)
```

amacati (changes requested):
I also reviewed the open PR from two weeks ago, which is contained here a second time. I'm also fine if you fix the comments here, we merge both together and discard the other PR.

lsy_drone_racing/envs/race_core.py
yufei4hua: That's true. I will apply the suggested changes, but split line 621, otherwise ruff complains.

lsy_drone_racing/envs/race_core.py
```diff
@@ -412,22 +414,21 @@ def _step(self, action: Array) -> tuple[dict[str, Array], float, bool, bool, dic
...
-        # https://github.com/jax-ml/jax/issues/29810
-        # Forcing a copy here is less efficient, but avoids the warning.
-        action = np.reshape(action, (self.sim.n_worlds, self.sim.n_drones, -1), copy=True)
+        assert action.shape == self.action_space.shape, (
+            f"Action shape mismatch: expected {self.action_space.shape}, got {action.shape}."
+        )
+        action = self._sanitize_action(
+            action, self.action_space.low, self.action_space.high, self.device
+        )
+        action = action.reshape((self.sim.n_worlds, self.sim.n_drones, -1))
```
amacati, line 431: Remove unnecessary reshape. Action is guaranteed to be correct after assert

amacati (commented):
One minor comment, then we can merge. Thanks a lot!

lsy_drone_racing/envs/race_core.py
yufei4hua: The reason I kept this is because action_space are built differently for different envs.
DroneRaceEnv: (n_act)
VecDroneRaceEnv: (n_worlds, n_act)
MultiDroneRaceEnv: (n_drones, n_act)
VecMultiDroneRaceEnv: (n_worlds, n_drones, n_act)
So we first assert for a specific environment, then still reshape to the sim required shape.

amacati: Ahh I forgot about the canonicalization, right!

PR #72, opened 2026-03-27
Add dev container support.

yufei4hua:
## Changes
- Dockerfile: pixi in docker.
- devcontainer.json: running pixi install in "postCreateCommand", adopted from official Pixi Docker configuration.
- updated docs to include WSL & Docker support.

## Function
- Simulation works well on Ubuntu, WSL2, and Docker Desktop (based on WSL2).
- Acados works well on all platforms.
- GPU can be enabled.
- On Windows (WSL2/Docker), rendering can only be achieved by software rendering, which is normally extremely slow (<20fps).
- I don't have a MacBook to test these functions.
- I didn't consider the deploy environment, since I suppose students should use our lab PCs.


mschuck-rai (approved):

amacati:
Thanks Yufei!

PR #74, opened 2026-03-27
Adding Multidrone Deployment Feature with Zenoh-ROS2

N0OBSTUDENT:
## Multi-drone real race support via host-client ROS2 architecture

- **Host-client split**: Replaces the monolithic `multi_deploy.py` with separate `deploy_host.py`
  and `deploy_client.py` scripts. The host manages all Crazyflie workers and race coordination;
  each client runs on its own compute unit and controls a single drone.

- **New environment classes**: Adds `RealMultiDroneRaceEnvHost` and `RealMultiDroneRaceEnvClient` —
  Gymnasium-compatible environments for the two sides of the host-client protocol.

- **ROS2 race communication layer**: New `RaceCommNode` in `ros_race_comm.py` handles all
  inter-process messaging (host-ready, race-start, client-state, clock calibration) using
  `lsy_race_msgs` ROS2 message types.

- **Clock synchronisation**: Clients calibrate their clock offset against the host via a ROS2
  service before the race starts.

- **Multi-drone controller wrappers**: Adds `AttitudeController` and `AttitudeMPC` subclasses
  that slice the per-drone observation from the joint multi-drone obs dict using the drone's `rank`.

- **Unknown observations default to NaN**: Considering that other drones could be out of mocap range, set the default observation to np.nan

- **Config updates**: `multi_level2.toml` and `multi_level0.toml` updated to fit the new
  architecture.


lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,367 @@
...
+        if self._client_state_pub:
+            try:
+                self._send_state_update(
+                    np.zeros(4 if self.control_mode == "attitude" else 13), stopped=True
+                )
+                time.sleep(0.1)  # allow the executor thread to flush the message before shutdown
+            except Exception as e:
+                logger.warning(f"Client {self.rank}: Could not send final stop message: {e}")
+        if self._comm:
+            self._comm.close()
```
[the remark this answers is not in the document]
N0OBSTUDENT: Added `self._ros_connector_own.close()` and `self._ros_connector_others.close()` upon closing of the client.

lsy_drone_racing/envs/utils.py
```diff
@@ -45,11 +45,20 @@ def load_track(track: ConfigDict) -> tuple[ConfigDict, ConfigDict, ConfigDict]:
...
-        k: np.array([drone.get(k) for drone in track.drones], dtype=np.float32)
-        for k in track.drones[0].keys()
+        "pos": drone_pos,
+        "quat": drone_quat,
+        "nominal_pos": drone_pos.copy(),
+        "nominal_quat": drone_quat.copy(),
     }
-    drones["quat"] = R.from_euler("xyz", drones["rpy"]).as_quat().astype(np.float32)
+
     return ConfigDict(gates), ConfigDict(obstacles), ConfigDict(drones)
```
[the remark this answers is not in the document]
N0OBSTUDENT: Restored to the original

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,816 @@
...
+        try:
+            cflib.crtp.init_drivers()
+            uri = f"radio://{self.rank}/{self.drone_channel}/2M/E7E7E7E7{self.drone_id:02X}"
+            PowerSwitch(uri).stm_power_cycle()
+            deadline = time.time() + 10.0
+            while time.time() < deadline:
+                if self.stop_event.is_set():
+                    return
+                time.sleep(0.05)
+
```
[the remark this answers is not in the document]
N0OBSTUDENT: Restored the initialization to the current commit of dev

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,816 @@
...
+                stop_event=stop_event,
+                failure_event=failure_event,
+            )
+            logger.debug(f"Spawned process for drone {rank} (PID: {process.pid})")
+
+        start_time = time.time()
+        while time.time() - start_time < 10.0:
+            if all(conn.ready_event.is_set() for conn in self._drone_connections.values()):
+                for conn in self._drone_connections.values():
+                    conn.connected = True
```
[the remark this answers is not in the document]
N0OBSTUDENT: Resolved by setting the sleeping time after powercycle to 2 seconds

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -31,6 +31,7 @@
 
 from lsy_drone_racing.envs.utils import gate_passed, load_track
 from lsy_drone_racing.utils.checks import check_drone_start_pos, check_race_track
+from lsy_drone_racing.utils.takeoff_barrier import TakeOffBarrier
```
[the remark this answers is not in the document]
N0OBSTUDENT: It is the prototype I coded using Barrier. Now the file is restored to the latest commit of branch dev.

lsy_drone_racing/control/attitude_controller.py
```diff
@@ -87,7 +88,7 @@ def compute_control(
             [r_des, p_des, y_des, t_des] as a numpy array.
         """
         t = min(self._tick / self._freq, self._t_total)
-        if t >= self._t_total:  # Maximum duration reached
+        if t >= 60.0:  # Maximum duration reached
```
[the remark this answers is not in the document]
N0OBSTUDENT: Restored the file to the latest commit of branch dev

scripts/deploy_client.py
```diff
@@ -0,0 +1,119 @@
...
+from lsy_drone_racing.utils import extract_config_for_rank, load_config, load_controller
+
+if TYPE_CHECKING:
+    from lsy_drone_racing.envs.real_race_env_client import RealMultiDroneRaceEnvClient
+
+logger = logging.getLogger(__name__)
+
+
+def main(
+    config: str = "multi_level2_single.toml",
```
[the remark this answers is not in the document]
N0OBSTUDENT: Forgot to restore it after testing. Changed it back to `multi_level2.toml`

lsy_drone_racing/control/attitude_controller.py
```diff
@@ -110,6 +111,7 @@ def compute_control(
         target_thrust[2] += self.drone_mass * self.g
 
         # Update z_axis to the current orientation of the drone
+        # z_axis = R.from_quat(obs["quat"]).as_matrix()[:, 2]
         z_axis = R.from_quat(obs["quat"]).as_matrix()[:, 2]
```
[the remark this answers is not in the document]
N0OBSTUDENT: Restored to the latest commit on branch dev

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,367 @@
...
+        self._host_ready_event = threading.Event()
+        self._race_started = False
+        self._race_start_time = 0.0
+        self._clock_offset = 0.0
+
+    def reset(self, *, seed: int | None = None, options: dict | None = None) -> tuple[dict, dict]:
+        """Reset the environment and wait for the host to signal readiness.
+
+        Sends dummy state messages at 10 Hz in the background so the host can detect
+        this client as ready. Blocks until :class:`HostReadyMessage` is received.
```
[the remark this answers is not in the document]
N0OBSTUDENT: Changed the docstring manually.

lsy_drone_racing/envs/utils.py
```diff
@@ -45,11 +45,20 @@ def load_track(track: ConfigDict) -> tuple[ConfigDict, ConfigDict, ConfigDict]:
     }
     obstacle_pos = np.array([o["pos"] for o in track.obstacles], dtype=np.float32)
```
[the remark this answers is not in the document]
N0OBSTUDENT: Restored the file to the latest commit of branch dev

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -172,6 +177,22 @@ def _reset(self, *, seed: int | None = None, options: dict | None = None) -> tup
             # Unlock thrust mode protection by sending a zero thrust command
             self.drone.commander.send_setpoint(0, 0, 0, 0)
 
+        if options.get("sync_start_barrier", True):
+            # Initialize barrier if not already created
+            barrier_config = options["takeoff_barrier"]
```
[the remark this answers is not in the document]
N0OBSTUDENT: It was for prototype using Barrier and is no longer needed. Restored the file to the latest commit of branch dev.

amacati:
There are merge conflicts at the moment. Is this ready for review?

N0OBSTUDENT:
> There are merge conflicts at the moment. Is this ready for review?

I mean not really... Marcel told me to have a draft PR so everybody knows what I am doing.

And I would like to ask for your advice, opinion and criticism on the structure and implementation. And I marked a few places with TODO, where I can not think of a better way to implement the functionalities.

N0OBSTUDENT:
And here is the test cases I would like to do, but after we made the structure clear.
https://docs.google.com/spreadsheets/d/1a_WvakzvkbI9FE5ZeVwlMI_acAK3nL8hTaW2EinpsPs/edit?usp=sharing

config/multi_level2.toml
ratheron, line 1: Please make sure the initial positions etc are equal between the levels

lsy_drone_racing/control/attitude_controller_multi.py
```diff
@@ -0,0 +1,59 @@
+"""This module implements an AttitudeController for quadrotor control.
+
+It utilizes the collective thrust interface for drone control to compute control commands based on
+current state observations and desired waypoints. The attitude control is handled by computing a
+PID control law for position tracking, incorporating gravity compensation in thrust calculations.
```
ratheron, line 5: Add documentation that it's just a wrapper around the default one

lsy_drone_racing/control/attitude_mpc.py
ratheron, line 1: Why is this necessary for this PR?

lsy_drone_racing/control/attitude_mpc_multi.py
```diff
@@ -0,0 +1,57 @@
+"""This module implements an example MPC using attitude control for a quadrotor.
+
+It utilizes the collective thrust interface for drone control to compute control commands based on
```
ratheron, line 3: Add documentation that it's just a wrapper around the default one. I think we don't need to provide an example on how to wrap another controller. This should be obvious from the other example. I think we should remove this entirely to not have too many options.

lsy_drone_racing/control/attitude_controller_multi.py
```diff
@@ -0,0 +1,59 @@
...
+            obs: The current observation of the environment. See the environment's observation space
+                for details.
+            info: Optional additional information as a dictionary.
+
+        Returns:
+            The orientation as roll, pitch, yaw angles, and the collective thrust
+            [r_des, p_des, y_des, t_des] as a numpy array.
+        """
+        obs = {
+            "pos": obs["pos"][self.rank],
```
ratheron, line 54: add a check for the dimensions so people get a proper error message when accidently using this for single-agent drone racing. Sth like
`assert pos.ndim == 2, f"Observation only has {pos.ndim} dimensions. This controller if for multi-agent only!"`

lsy_drone_racing/control/attitude_rl_multi.py
ratheron, line 1: Same as MPC. I don't think we need this.

lsy_drone_racing/envs/real_race_env.py
ratheron, line 1: Nice addition to prevent the crash. However, there will be a new error now, no? Maybe we can additionally catch the timeout of the ros_connector too?

lsy_drone_racing/utils/ros.py
ratheron, line 1: See comment on real_race_env

lsy_drone_racing/utils/utils.py
ratheron, line 1: Looks elegant, nice. Do we extract the whole dict and remove all infos on other drones, or do we only extract the info for the drone with the specific rank?
Can we add tests for this, please?

ros_ws/src/lsy_race_msgs/msg/ClientState.msg
```diff
@@ -0,0 +1,6 @@
+int32 drone_rank
+float64[] action
+float64 elapsed_time
+float64 timestamp
+bool stopped
```
ratheron, line 5: stopped is controller.finished? We should keep the naming the same. Maybe controller_finished?

ros_ws/src/lsy_race_msgs/msg/RaceStart.msg
```diff
@@ -0,0 +1,3 @@
+float64 elapsed_time
+float64 timestamp
+bool finished
```
ratheron, line 3: With the other comment, maybe we rename this tho race_finished?

pyproject.toml
ratheron, line 1: Since you have switched to Zenoh and we now always need another terminal running a central communication server (afaik), add that to the docs. In general, this PR should also add the docs on how to use the code - at least a rough version - because as I see there are also new tasks (nice idea btw)

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+        self.last_drone_pos = np.zeros((n_drones, 3), dtype=np.float32)
+        self.taken_off = False
+
+    def reset(self, last_drone_pos: NDArray[np.float32]):
+        """Reset all dynamic fields and seed last drone positions."""
+        self.target_gate[:] = 0
+        self.gates_visited[:] = False
+        self.obstacles_visited[:] = False
+        self.last_drone_pos[:] = last_drone_pos
+        self.taken_off = False
```
ratheron, line 52: Why not reuse/expand EnvData directly? It just adds the take_off flag, no? This might be useful later for the regular env, too.

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+class RealMultiDroneRaceEnvClient(Env):
+    """Client-side Gymnasium environment for multi-drone racing.
+
+    Runs on each drone's computing unit. Receives host coordination messages via ROS2,
+    computes observations and gate tracking locally, and forwards actions to the host
+    which relays them to the physical drone.
+
+    Observation space:
+        A dictionary containing the state of all drones in the race, mirroring
+        :class:`lsy_drone_racing.envs.multi_drone_race.MultiDroneRaceEnv`.
```
ratheron, line 64: This PR should delete MultiDroneRaceEnv, since it's no longer needed, right?

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+    def reset(self, last_drone_pos: NDArray[np.float32]):
+        """Reset all dynamic fields and seed last drone positions."""
+        self.target_gate[:] = 0
+        self.gates_visited[:] = False
+        self.obstacles_visited[:] = False
+        self.last_drone_pos[:] = last_drone_pos
+        self.taken_off = False
+
+
+class RealMultiDroneRaceEnvClient(Env):
```
ratheron, line 55: Shouldn't we inherit RaceCoreEnv, as done in MultiDroneRaceEnv to reuse capabilities where possible?

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+        """Open ROS connectors for own drone (estimator) and others (TF)."""
+        self._ros_connector_own = ROSConnector(
+            estimator_names=[self.drone_name],
+            cmd_topic=f"/drones/{self.drone_name}/command",
+            timeout=10.0,
+        )
+
+        other_names = [n for i, n in enumerate(self.drone_names) if i != self.rank]
+        if other_names:
+            self._ros_connector_others = ROSConnector(tf_names=other_names, timeout=10.0)
```
ratheron, line 318: Why not have one ros_connector? One should be able to provide both, raw tfs and estimates, just not of the same drone.

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+            seed: Unused in real environments.
+            options: Unused in real environments.
+
+        Returns:
+            Initial observation and info dictionaries.
+
+        Raises:
+            TimeoutError: If the host does not respond within 120 seconds.
+        """
+        if options and options.get("real_track_objects", False):
```
ratheron, line 137: If options in unused, why are we using options.real_track_objects?

scripts/deploy_client.py
```diff
@@ -0,0 +1,117 @@
...
+from lsy_drone_racing.utils import extract_config_for_rank, load_config, load_controller
+
+if TYPE_CHECKING:
+    from lsy_drone_racing.envs.real_race_env_client import RealMultiDroneRaceEnvClient
+
+logger = logging.getLogger(__name__)
+
+
+def main(
+    config: str = "multi_level2.toml", controller: str | None = None, drone_rank: int | None = None
```
ratheron, line 33: Isn't there a more direct way for passing the own drone? Passing the drone name as for example `cf20` is easier for the user, no? And it should be easy to get the rank from the config.

scripts/deploy_client.py
```diff
@@ -0,0 +1,117 @@
...
+    """Deploy and run a client for multi-drone racing.
+
+    Args:
+        config: Configuration file name, assumed to be in ``config/``.
+        controller: Controller file name in ``lsy_drone_racing/control/``, or ``None``
+            to use the value from the config file.
+        drone_rank: Rank of this drone in the multi-drone setup. Required.
+    """
+    if drone_rank is None:
+        raise ValueError("drone_rank must be specified")
```
ratheron, line 44: rather use assert

scripts/deploy_host.py
```diff
@@ -0,0 +1,62 @@
...
+
+def main(config: str = "multi_level2.toml"):
+    """Deploy and run the race host.
+
+    Args:
+        config: Configuration file name, assumed to be in ``config/``.
+    """
+    rclpy.init()
+    config_obj = load_config(Path(__file__).parents[1] / "config" / config)
+    deploy = config_obj.deploy
```
ratheron, line 34: in deploy client, you use config_obj.deploy everywhere

scripts/deploy_client.py
```diff
@@ -0,0 +1,117 @@
...
+            obs, reward, terminated, truncated, info = env.step(action)
+            controller_finished = controller.step_callback(
+                action, obs, reward, terminated, truncated, info
+            )
+            if terminated or truncated or controller_finished:
+                logger.debug(
+                    f"Client {drone_rank}: Episode finished "
+                    f"(terminated={terminated}, "
+                    f"truncated={truncated}, "
+                    f"finished={controller_finished})"
```
ratheron, line 86: cleaner way for `f"terminated={terminated}"` is `f"{terminated=}"`

lsy_drone_racing/envs/real_race_env_client.py
ratheron, line 52: We should reuse EnvData from RealRaceCoreEnv

lsy_drone_racing/envs/real_race_env_client.py
ratheron, line 318: Since we only get the tfs of the other drones, we can get that information from the own ros connector. Thus, we only need one connector.

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+        quat = np.full((self.n_drones, 4), np.nan, dtype=np.float32)
+        vel = np.full((self.n_drones, 3), np.nan, dtype=np.float32)
+        ang_vel = np.full((self.n_drones, 3), np.nan, dtype=np.float32)
+        pos[self.rank] = self._ros_connector_own.pos[self.drone_name]
+        quat[self.rank] = self._ros_connector_own.quat[self.drone_name]
+        vel[self.rank] = self._ros_connector_own.vel[self.drone_name]
+        ang_vel[self.rank] = self._ros_connector_own.ang_vel[self.drone_name]
+        if self._ros_connector_others is not None:
+            for i, name in enumerate(self.drone_names):
+                if i != self.rank:
```
ratheron, line 367: Cleaner:
`if i == self.rank:
        continue`

lsy_drone_racing/envs/real_race_env_client.py
ratheron, line 55: Why not inherit from RealRaceCoreEnv? Much of the functionality should be identical.

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+
+if TYPE_CHECKING:
+    from ml_collections import ConfigDict
+    from numpy.typing import NDArray
+    from rclpy.publisher import Publisher
+
+logger = logging.getLogger(__name__)
+
+
+class CrazyflieWorker:
```
ratheron, line 50: This class should also inherit RealRaceCoreEnv. Or better: This class is used by the real race core env. In that case, we should move it somewhere else.

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+            return (
+                self.stop_event.is_set()
+                or self.failure_event.is_set()
+                or self.connection_lost_event.is_set()
+            )
+
+        # TODO: Really bad, but I do not know a better way to do it.
+        # The motivation is that the Worker should
+        # NEITHER throw exceptions NOR receive interrupt signals.
+        # Otherwise it would be very chaotic.
```
ratheron, line 361: How about
```
steps = [
    self._connect_drone,
    self._crazyflie_reset,
    self._init_ros_connector,
    self._init_ros_comm,
    self.ready_event.set
]

for step in steps:
    if early_stop():
        return
    step()
```

ratheron (changes requested):
Generally nice implementation with proper safe guards. Thank you for the work!

lsy_drone_racing/control/attitude_controller_multi.py
amacati: Also explicitly mention that it is meant for multi-drone racing. It's in the name, but should also be explained.

lsy_drone_racing/control/attitude_mpc_multi.py
amacati: +1 for only keeping one example

lsy_drone_racing/control/attitude_rl_multi.py
amacati: +1

ros_ws/src/lsy_race_msgs/msg/ClientState.msg
amacati, line 1: General comment on the ros stuff: I am not happy at all with having a `ros_ws` in a python package. Ros packages should go into whatever catkin workspace is already there on the system, not be included inside a Python package. How do other packages handle this (apart from crazyswarm)? Do we need these messages? Can't we build them out of existing ones?

scripts/deploy_host.py
```diff
@@ -0,0 +1,62 @@
...
+            check_objects=deploy.real_track_objects and deploy.check_race_track,
+            check_drones=deploy.check_drone_start_pos,
+        )
+        host.connect_drones()
+        logger.info("Drones connected, starting main loop...")
+        host.host_main_loop()
+    except KeyboardInterrupt:
+        logger.info("Interrupted, shutting down...")
+    except Exception as e:
+        logger.error(f"Host encountered error: {e}", exc_info=True)
```
amacati, line 49: Why do we need a bare except here? That's an absolute last resort. Finally triggers anyways

scripts/deploy_client.py
amacati, line 33: How do we ensure the same configs are selected on all clients? Do the settings need to match somehow, or are heterogeneous settings even desirable?

scripts/deploy_client.py
amacati: None does not make sense in the first place. Why type hint a None if None is not possible? Why default to a value that throws an error?

lsy_drone_racing/envs/real_race_env_client.py
amacati: +1, we should try to keep things as close as possible.

lsy_drone_racing/envs/real_race_env_client.py
amacati: That's the sim class, and it should stay

amacati (commented):
Left some comments on the surrounding changes. 

TODO @amacati review `ros_race_comm`, `real_race_env_client`, `real_race_host`.

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+            self.gates.pos, self.gates.quat, self.obstacles.pos = track_poses(
+                self.n_gates, self.n_obstacles
+            )
+
+        if self._ros_connector_own is None:
+            self._init_ros_connectors()
+        if self._comm is None:
+            self._init_comm()
+
+        current_pos, _, _, _ = self._get_all_drone_states()
```
amacati, line 147: Nit: Avoid getter names

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+        self.data.reset(current_pos)
+
+        logger.debug("Environment reset complete")
+        return self.obs(), self.info()
+
+    def lock_until_race_start(self, timeout: float = 60.0):
+        """Sends dummy messages at the control frequency (``self.freq`` Hz) until the race starts.
+
+        After receiving host ready message, the client will calibrate the clock offset and \
+            until the host broadcasts :class:`RaceStartMessage`.
```
amacati, line 157: Why the `\`?

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+    def step(self, action: NDArray) -> tuple[dict, float, bool, bool, dict]:
+        """Perform a control step: update gate tracking, check bounds, and send the action.
+
+        Args:
+            action: Control action for this drone.
+
+        Returns:
+            Observation, reward (always 0.0), terminated, truncated (always False), info.
+        """
+        drone_pos, _, _, _ = self._get_all_drone_states()
```
amacati, line 209: Still prefer not to have `get_xxx` methods. `all_drone_states()` should be enough

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+        self.data.target_gate += np.asarray(passed)
+        self.data.target_gate[self.data.target_gate >= self.n_gates] = -1
+        self.data.last_drone_pos[...] = drone_pos
+        self.data.taken_off |= drone_pos[self.rank, 2] > 0.1
+
+        terminated = bool(self.data.target_gate[self.rank] == -1)
+        if np.any(
+            (self.pos_limit_low > drone_pos[self.rank])
+            | (drone_pos[self.rank] > self.pos_limit_high)
+        ):
```
amacati, line 231: This condition should be spread out for readability

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def close(self):
+        """Send a final stop message and close all ROS connections."""
+        logger.info("Closing environment...")
+        if self._client_state_pub:
+            try:
+                self._send_state_update(
+                    np.zeros(4 if self.control_mode == "attitude" else 13), stopped=True
+                )
+                time.sleep(0.1)  # allow the executor thread to flush the message before shutdown
```
amacati, line 277: Can we ensure this instead of just waiting? I.e. send something with flush=True?

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+            except Exception as e:
+                logger.warning(f"Could not send final stop message: {e}")
```
amacati, line 279: No bare excepts except for exceptional requirements.

Rule of thumb, if we have a bare except, I would expect about 5 lines of comments above why this is **absolutely** necessary.

lsy_drone_racing/envs/real_race_env_client.py
amacati: Second that, do we need a second connector?

lsy_drone_racing/envs/real_race_env_client.py
```diff
@@ -0,0 +1,372 @@
...
+        quat[self.rank] = self._ros_connector_own.quat[self.drone_name]
+        vel[self.rank] = self._ros_connector_own.vel[self.drone_name]
+        ang_vel[self.rank] = self._ros_connector_own.ang_vel[self.drone_name]
+        if self._ros_connector_others is not None:
+            for i, name in enumerate(self.drone_names):
+                if i != self.rank:
+                    pos[i] = self._ros_connector_others.pos.get(name, np.nan)
+                    quat[i] = self._ros_connector_others.quat.get(name, np.nan)
+                    # vel[i] = self._ros_connector_others.vel.get(name, np.nan)
+                    # ang_vel[i] = self._ros_connector_others.ang_vel.get(name, np.nan)
```
amacati, line 371: Remove comments

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+                    self.failure_event.set()
+                    self.connection_lost_event.set()
+
+            self.drone.fully_connected.add_callback(on_connected)
+            self.drone.connection_failed.add_callback(on_connection_failed)
+            self.drone.connection_lost.add_callback(on_connection_lost)
+            self.drone.open_link(uri)
+
+            start_time = time.time()
+            while time.time() - start_time < 10.0:
```
amacati, line 224: Avoid magic numbers

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+                    f"Timed out waiting for drone {self.drone_id} on channel {self.drone_channel}."
+                )
+                return
+
+            self.logger.info(f"Connected to {uri}")
+        except Exception:
+            # If anything goes wrong during connection, set shared failure event to notify the host
+            self.logger.error(f"Exception while connecting to drone {self.drone_id}", exc_info=True)
+            self.failure_event.set()
+            return
```
amacati, line 252: Same story about bare excepts

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+            self._comm = RaceCommNode(f"lsy_race_worker_{self.rank}")
+            self._sub = self._comm.node.create_subscription(
+                ClientState,
+                f"lsy_drone_racing/client/drone_{self.rank}/state",
+                self._on_client_state,
+                10,
+            )
+        except Exception as e:
+            self.logger.error(f"Failed to initialize ROS communication: {e}", exc_info=True)
+            self.failure_event.set()
```
amacati, line 266: Bare except

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+        self.logger.info(f"Initializing ROS connector for {self.drone_name}...")
+        try:
+            self._ros_connector = ROSConnector(
+                estimator_names=[self.drone_name],
+                cmd_topic=f"/drones/{self.drone_name}/command",
+                timeout=10.0,
+            )
+        except Exception as e:
+            self.logger.error(f"Failed to initialize ROS connector for {self.drone_name}: {e}")
+            self.failure_event.set()
```
amacati, line 279: Bare except

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+    """
+
+    _initialized: bool = False
+    _num_drones: int = 0
+    _config: ConfigDict | None = None
+    _comm: RaceCommNode | None
+    _host_ready_pub: Publisher | None
+    _race_start_pub: Publisher | None
+
+    def __init__(self, config: ConfigDict):
```
amacati, line 455: No initialization via ConfigDict as the only argument. We should be explicit over implicit. With this, we have no idea what the arguments to this class really are.

lsy_drone_racing/utils/ros_race_comm.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Args:
+        timestamp: Time the message was sent.
+        clock_offset: Calibrated offset (host_time − client_time) in seconds.
+            Zero when called on the host side (timestamps are already in host time).
+
+    Returns:
+        Estimated one-way latency in milliseconds.
+    """
+    return (time.time() - timestamp - clock_offset) * 1000
```
amacati, line 66: Do we really need this function? It's 11 lines for a one-liner that is very readable

lsy_drone_racing/utils/ros_race_comm.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _hook(args: threading.ExceptHookArgs) -> None:
+        if args.exc_type in (ExternalShutdownException, KeyboardInterrupt) or (
+            args.exc_type.__name__ == "RCLError"
+        ):
+            logger.debug(f"Thread '{args.thread.name}' stopped (shutdown)")
+        else:
+            _original(args)
+
+    threading.excepthook = _hook
```
amacati, line 45: Function definition before imports are finished

lsy_drone_racing/utils/ros_race_comm.py
```diff
@@ -0,0 +1,150 @@
...
+    if not client.wait_for_service(timeout_sec=timeout):
+        raise TimeoutError(f"Calibration service not available after {timeout}s")
+    offsets = []
+    for _ in range(n):
+        t_send = time.time()
+        future = client.call_async(CalibrateClock.Request())
+        while not future.done():
+            time.sleep(0.0001)  # Don't busy-wait too aggressively
+        t_recv = time.time()
+        offsets.append(future.result().host_timestamp - (t_send + t_recv) / 2)
```
amacati, line 101: Why not use the blocking version?

lsy_drone_racing/utils/ros_race_comm.py
```diff
@@ -0,0 +1,150 @@
...
+                    raise
+
+        self._thread = threading.Thread(target=_spin, daemon=True, name=f"spin-{name}")
+        self._thread.start()
+        logger.debug(f"RaceCommNode '{name}' started")
+
+    @property
+    def node(self) -> Node:
+        """The underlying rclpy node."""
+        return self._node
```
amacati, line 144: Why do we need the property then?

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+        def early_stop() -> bool:
+            # If the connection lost
+            # or any of the other workers reported a failure
+            # or we were asked to stop during connection,
+            # just exit
+            return (
+                self.stop_event.is_set()
+                or self.failure_event.is_set()
+                or self.connection_lost_event.is_set()
+            )
```
amacati, line 356: Why the inline definition?

lsy_drone_racing/envs/real_race_host.py
amacati, line 50: Could we turn this into one function instead? We only execute run() anyways

lsy_drone_racing/envs/real_race_host.py
```diff
@@ -0,0 +1,812 @@
...
+    failure_event: mp.synchronize.Event | None = None
+    connected: bool = False
+
+
+class RealRaceHost:
+    """Base class for multi-drone race hosts.
+
+    Subclasses implement :meth:`load_config`, :meth:`connect_drones`,
+    :meth:`host_main_loop`, and :meth:`close` for a specific drone platform.
+    """
```
amacati, line 446: Why do we subclass this? We're only running crazyflies for now. Also, if anything, this should then be a protocol

lsy_drone_racing/utils/ros_race_comm.py
amacati, line 45: This is fairly complex. Was this a major issue during testing on the real hardware, or why was this included?

lsy_drone_racing/control/attitude_controller_multi.py
N0OBSTUDENT: Added the assertion

lsy_drone_racing/control/attitude_mpc_multi.py
N0OBSTUDENT: Okay, I have modified the multi drone attitude mpc and deleted other multi drone controller

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: It is meant to fix the unexpected error. Now it is completely idendical to commit 8ecfe76 . Maybe you could revert this change in the PR

ros_ws/src/lsy_race_msgs/msg/ClientState.msg
N0OBSTUDENT: Modified in https://github.com/N0OBSTUDENT/drone_racing_msgs/tree/real_race and once I tested them out it will be merged into https://github.com/rducrist/drone_racing_msgs

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: Done. Imported from real_race_env

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: As discussed last time we will keep it as it is right now...

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: I did not know we could do that, now it's one.

scripts/deploy_client.py
N0OBSTUDENT: Getting the rank from the toml file is a reverse checking process, isn't it? If we pass the drone name, we have to iterate through the drone configuration list decide the rank.

scripts/deploy_client.py
N0OBSTUDENT: I would rather change it to assert.
The rank is a required field, but I do not want to change the order of arguments compared to the single drone racing. If I set the drone rank to some feasible default value, it will run silently and send actions to probably wrong drone if any of the students forget to set the drone rank upon launching the client.

scripts/deploy_host.py
N0OBSTUDENT: Okay, I will unify the style. I would prefer dropping this line

scripts/deploy_client.py
N0OBSTUDENT: Okay, didn't know that

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: Imported EnvData from real_race_env.py

lsy_drone_racing/envs/real_race_env.py
N0OBSTUDENT: I have reverted that. Sticked to the commit https://github.com/utiasDSL/lsy_drone_racing/commit/8ecfe7643b17919085db49bfc6c2a331c18c0a4a

ros_ws/src/lsy_race_msgs/msg/RaceStart.msg
N0OBSTUDENT: Will be solved in https://github.com/utiasDSL/lsy_drone_racing/pull/74#discussion_r3144236992

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: That was a mistake. I did not add track object pose update on client side before. After I added the feature I did not update the docstring

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: Modified in https://github.com/N0OBSTUDENT/lsy_drone_racing/commit/6fdd0786466eac72eb847ff0f49c45f8dace485e

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: No longer needed now

ros_ws/src/lsy_race_msgs/msg/ClientState.msg
N0OBSTUDENT: Now I am cloning the message from an external repository, following the style of mocap pacakge. We will be sharing the pacakge with simulation on https://github.com/rducrist/drone_racing_msgs

scripts/deploy_host.py
N0OBSTUDENT: The problem now is that multiple processes may throw tons of messages. I was trying to make the belonging of the message clearer

scripts/deploy_client.py
N0OBSTUDENT: Currently there is no way we could ensure that... I suppose the setting do not need to match perfectly for every client. Only the track and drone settings need to be the same

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: Deleted... It is not intended

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: Modified in https://github.com/N0OBSTUDENT/lsy_drone_racing/commit/bf9666145bbad471860dd6b1049a23cf4325dbd8

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: I was encountering some issues with ros publisher closing too early before the last message is actually sent. I will remove it now.

lsy_drone_racing/envs/real_race_env_client.py
N0OBSTUDENT: I will test whether the problem still exists rightnow. By then I was encountering issues with ros publisher closing too fast before the last message is sent

lsy_drone_racing/envs/real_race_host.py
N0OBSTUDENT: But this is added in case the CrazyFlies Drone object throws any exception, which, has plenty of variants.
I am now not using the try-except here anymore, but I am keeping the bare except in the run function.

lsy_drone_racing/envs/real_race_host.py
N0OBSTUDENT: It is doable, but the function will be as long as a tiolet paper roll

lsy_drone_racing/envs/real_race_host.py
N0OBSTUDENT: Modified in https://github.com/N0OBSTUDENT/lsy_drone_racing/commit/b671a7422292dd9dccc98c36f2676ba60c5866f3

lsy_drone_racing/utils/ros_race_comm.py
N0OBSTUDENT: Modified in [N0OBSTUDENT@bbd9c31](https://github.com/N0OBSTUDENT/lsy_drone_racing/commit/bbd9c31920460e30465cb15d5148411b6306a3d2)
I will test if it works as expected

lsy_drone_racing/utils/ros_race_comm.py
N0OBSTUDENT: That's my mistake. Changed it in https://github.com/N0OBSTUDENT/lsy_drone_racing/commit/2ed0147ff70a7a474d2206df94c9eac5b5ed99e3

lsy_drone_racing/envs/real_race_host.py
N0OBSTUDENT: My initial is that we could have different implementation of the Host with common communiation between the client and the host. So I had communicators initialized in RealRaceHost...

lsy_drone_racing/control/attitude_mpc.py
N0OBSTUDENT: That's my mistake, will be retracted in the following PR

PR #76, opened 2026-04-07
Add render() lifecycle and race-core unit tests

Yuming-Lee24:
## Summary

This branch expands test coverage for `lsy_drone_racing/envs/race_core.py` in two layers:

**Render lifecycle smoke tests** (integration)
- `test_render` covers `DroneRacing-v0`: renders right after reset, mid-episode (5 steps), and after forcing the drone into a terminal state via `target_gate = -1`.
- `test_render_multi_drone` covers `MultiDroneRacing-v0`: renders after reset, mid-episode, with one drone disabled (exercises `_warp_disabled_drones` on a partial mask while autoreset stays inactive), and after forcing all drones terminal (exercises the autoreset path).
- Both are gated by a `skip_if_headless` marker so CI without a display does not fail.
- These are the first tests that exercise `RaceCoreEnv.render()`, which previously had no coverage.

**Race-core unit tests** (`tests/unit/envs/test_race_core.py`, new file)
- **close**: `test_close_after_reset`, `test_close_without_reset`.
- **obs**: `test_obs_structure_and_initial_values`, `test_obs_returns_nominal_when_out_of_sensor_range`, `test_obs_returns_real_pose_when_in_sensor_range` (the last two pin down the sensor-range fog-of-war in `_obs`).
- **reward**: `test_reward_zero_during_race`, `test_reward_minus_one_when_course_finished` (both branches of the sparse reward).
- **terminated**: `test_terminated_false_after_reset`, `test_terminated_true_when_target_gate_negative`.
- **truncated**: `test_truncated_false_after_reset`, `test_truncated_on_timeout_does_not_terminate` (also asserts that timeout and termination are independent signals, not equivalent).
- **`_disabled_drones`**: `test_disabled_drones_nominal_not_disabled`, `test_disabled_drones_out_of_bounds`, `test_disabled_drones_on_contact` — called directly with crafted inputs so each disable branch is exercised in isolation.
- **`_step_env` / gate-pass detection**: `test_gate_pass_increments_target_gate` synthetically places `last_drone_pos` and `sim.data.states.pos` 5 cm on either side of the current target gate along its local +x axis, then calls `_step_env` directly (bypassing physics) and asserts `target_gate` increments from 0 to 1. Paired with `test_gate_not_passed_without_crossing` as the negative case.

The unit tests follow the existing `env.unwrapped.data.replace(...)` pattern to craft terminal/timeout/gate-pass states without relying on a full flight, so they are deterministic and fast (~12 s for the full file).

## Coverage notes

Directly tested in `test_race_core.py`: `close`, `obs`, `reward`, `terminated`, `truncated`, `_disabled_drones`, `_step_env`.
Indirectly exercised via setup/step: `__init__`, `_reset`, `_step`, `_obs`, `_truncated`, `_reset_env_data`, `_setup_sim`, `_load_track_into_sim`, `EnvData.create`, `build_action_space`, `build_observation_space`.
Still untested by this PR: `apply_action` / `_sanitize_action` (action clipping & control-mode dispatch), `info`, `drone_mass`, `rng_spec2fn`, `build_dynamics_disturbance_fn`. These are candidates for a follow-up.

## Test plan

- [x] `pytest tests/integration/test_envs.py::test_render -v`
- [x] `pytest tests/integration/test_envs.py::test_render_multi_drone -v`
- [x] `pytest tests/unit/envs/test_race_core.py -v` (16 passed)
- [x] `ruff check tests/unit/envs/test_race_core.py` (clean)
- [x] CI green on headless runner (verify `skip_if_headless` still skips render tests)

### DIFF
--- a/tests/integration/test_envs.py
+++ b/tests/integration/test_envs.py
@@ -1,4 +1,3 @@
-import os
 from pathlib import Path
 
 import gymnasium
@@ -10,11 +9,7 @@ from drone_models import available_models
 import lsy_drone_racing  # noqa: F401, environment registrations
 from lsy_drone_racing.envs.drone_race import DroneRaceEnv
 from lsy_drone_racing.utils import load_config
-
-skip_if_headless = pytest.mark.skipif(
-    os.environ.get("DISPLAY") is None,
-    reason="DISPLAY is not set, skipping test in headless environment",
-)
+from tests.conftest import skip_if_headless
 
 CONFIG_FILES = {
     "DroneRacing-v0": ["level0.toml", "level1.toml", "level2.toml"],
--- a/tests/integration/test_envs.py
+++ b/tests/integration/test_envs.py
@@ -187,3 +187,86 @@ def test_vector_envs_randomization(config_file: str):
     )
 
     env.close()
+
+
+@pytest.mark.integration
+def test_render():
+    """Smoke test: env.render() runs at every lifecycle stage for DroneRacing-v0."""
+    config = load_config(Path(__file__).parents[2] / "config" / "level0.toml")
+    env = gymnasium.make(
+        "DroneRacing-v0",
+        freq=config.env.freq,
+        sim_config=config.sim,
+        sensor_range=config.env.sensor_range,
+        control_mode="state",
+        track=config.env.track,
+        disturbances=config.env.get("disturbances"),
+        randomizations=config.env.get("randomizations"),
+        seed=config.env.seed,
+    )
+    # Right after reset
+    env.reset()
+    env.render()
+    # Mid-episode: render after several steps
+    for _ in range(5):
+        env.step(env.action_space.sample())
+        env.render()
+
+    # Force terminal state
+    data = env.unwrapped.data
+    env.unwrapped.data = data.replace(target_gate = data.target_gate.at[...].set(-1))
+    env.step(env.action_space.sample())
+    assert jp.all(env.unwrapped.data.disabled_drones), "drone not actually disabled"
+    env.render()
+    env.close()
+
+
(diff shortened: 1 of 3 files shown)


tests/integration/test_envs.py
```diff
@@ -10,6 +11,11 @@
 from lsy_drone_racing.envs.drone_race import DroneRaceEnv
 from lsy_drone_racing.utils import load_config
 
+skip_if_headless = pytest.mark.skipif(
+    os.environ.get("DISPLAY") is None,
+    reason="DISPLAY is not set, skipping test in headless environment",
+)
+
```
ratheron: Please move this to conftest, as done in [Crazyflow](https://github.com/utiasDSL/crazyflow/blob/main/tests/conftest.py#L41-L45)

tests/unit/envs/test_race_core.py
```diff
@@ -0,0 +1,306 @@
...
+    contacts = env.unwrapped.sim.contacts()
+    new_data = RaceCoreEnv._step_env(
+        data,
+        env.unwrapped.sim.data.states.pos,
+        env.unwrapped.sim.mjx_data.mocap_pos,
+        env.unwrapped.sim.mjx_data.mocap_quat,
+        contacts,
+    )
+    assert int(np.asarray(new_data.target_gate[0, 0])) == 0
+    env.close()
```
ratheron, line 306: We should also check if the gate is not passed when going through it in reverse. The second test can be written as passing in the plane, but next to the gate. This should also not be counted as a gate pass.

ratheron (changes requested):
Overall nice PR. Thanks! Please add the two changes, then we can merge from my side.

PR #77, opened 2026-04-10
Enable n controllers to run in multisim

rducrist:
This PR updates the multi-drone simulation to support one controller per drone instead of broadcasting one action to all drones. 

### Changes: 

- load one controller per drone in `multi_sim.py`
- support CLI override via `--controllers=a.py,b.py`
- add multi-controller attitude/MPC wrappers using `info["rank"]`
- log per-drone finish times and completion state


lsy_drone_racing/control/attitude_controller_multi.py
```diff
@@ -0,0 +1,59 @@
+"""This module implements an AttitudeController for quadrotor control.
+
+It utilizes the collective thrust interface for drone control to compute control commands based on
+current state observations and desired waypoints. The attitude control is handled by computing a
+PID control law for position tracking, incorporating gravity compensation in thrust calculations.
+
+The waypoints are generated using cubic spline interpolation from a set of predefined waypoints.
+Note that the trajectory uses pre-defined waypoints instead of dynamically generating a good path.
+"""
```
ratheron: Add documentation that it's just a wrapper around the default one. Also explicitly mention that it is meant for multi-drone racing. It's in the name, but should also be explained.
See [this comment](https://github.com/utiasDSL/lsy_drone_racing/pull/74#discussion_r3032342390)

lsy_drone_racing/control/attitude_mpc_multi.py
```diff
@@ -0,0 +1,57 @@
+"""This module implements an example MPC using attitude control for a quadrotor.
+
+It utilizes the collective thrust interface for drone control to compute control commands based on
+current state observations and desired waypoints.
+
+The waypoints are generated using cubic spline interpolation from a set of predefined waypoints.
+Note that the trajectory uses pre-defined waypoints instead of dynamically generating a good path.
+"""
```
ratheron: See comment on the attitude controller

scripts/multi_sim.py
```diff
@@ -112,22 +134,35 @@ def simulate(
                         if not e.args[0].startswith("No known conversion for Jax type"):
                             raise e
```
ratheron, line 141: Do we still need the try?

scripts/multi_sim.py
```diff
@@ -112,22 +134,35 @@ def simulate(
...
-    finished = gates_passed == -1
-    logger.info((f"Flight time (s): {curr_time}\nDrones finished: {finished}\n"))
+    n_gates = len(config.env.track.gates)
+    gates_passed = np.where(gates_passed == -1, n_gates, gates_passed)
+    finished = gates_passed == n_gates
+    logger.info(
+        "Flight time (s): %s\nFinished: %s\nGates passed: %s\n",
+        finish_times.tolist(),
+        finished.tolist(),
+        gates_passed.tolist(),
```
ratheron: I think showing it per controller is easier to comprehend. Also, why not use f strings? Maybe display a full table?

ratheron (changes requested):
Thanks for that PR! I left some comments for some minor changes

scripts/multi_sim.py
amacati: No, this has been fixed upstream in gymnasium

scripts/multi_sim.py
rducrist: image

Like this?

config/multi_level0.toml
```diff
@@ -4,9 +4,9 @@
 # | :-----------------: | :-----------------------: | :-------------------------: | :--------------------: | :---------------: |
 # | `level0.toml`       | *No*                      | *No*                        | *No*                   | Perfect knowledge |
 [[controller]]
-file = "state_controller.py" # Put your controller file name here. Specifying a controller as argument to scripts will override this setting. Controllers are located in `lsy_drone_racing/control/`
+file = "attitude_mpc_multi.py" # Put your controller file name here. Specifying a controller as argument to scripts will override this setting. Controllers are located in `lsy_drone_racing/control/`
```
ratheron, line 7: Why don't we race a state controller against an attitude controller. Then we would automatically test that both control interfaces work together. Or are different control interface not possible atm?

lsy_drone_racing/control/attitude_controller.py
```diff
@@ -51,8 +51,10 @@ def __init__(self, obs: dict[str, NDArray[np.floating]], info: dict, config: dic
         self.g = 9.81
 
         # Same waypoints as in the position controller. Determined by trial and error.
+        start_position = obs.get("pos")
         waypoints = np.array(
             [
+                start_position,
                 [-1.5, 0.75, 0.05],
```
ratheron, line 58: Shouldnt we remove the old first point then? This is only 5cm above the floor.

lsy_drone_racing/control/attitude_mpc.py
```diff
@@ -184,8 +184,10 @@ def __init__(self, obs: dict[str, NDArray[np.floating]], info: dict, config: dic
         self._T_HORIZON = self._N * self._dt
 
         # Same waypoints as in the trajectory controller. Determined by trial and error.
+        start_position = obs.get("pos")
         waypoints = np.array(
             [
+                start_position,
                 [-1.5, 0.75, 0.05],
```
ratheron, line 191: Same as in the attitude controller, this point should be removed imo. Also for all other baseline controllers.

scripts/multi_sim.py
ratheron: @rducrist please remove the try except block. Only keep the render, as in the other sim.py scrip[t

scripts/multi_sim.py
```diff
@@ -81,28 +87,50 @@ def simulate(
...
-
-            action = controller.compute_control(obs, info)
-            action = np.array([action] * n_drones * n_worlds, dtype=np.float32)
-            action[1, 0] += 0.2
-            obs, reward, terminated, truncated, info = env.step(action)
-            done = terminated | truncated
-            # Update the controller internal state and models.
-            controller.step_callback(action, obs, reward, terminated, truncated, info)
-            # Add up reward, collisions
+            ranked_infos = [_info_with_rank(info, rank) for rank in range(n_drones)]
```
ratheron: why not directly do 
`ranked_infos = [{**info, "rank": rank} for rank in range(n_drones)]`

ratheron (changes requested):
Left some minor comments. I think we are close to merging this PR!

config/multi_level0.toml
rducrist: Afaik no. `_build_action_space()` in `race_core` currently limits to either or.

PR #84, opened 2026-04-28
Update to new orga name

amacati:
@ratheron I think we have to update this asap to not break the challenge. Can you check if the leaderboard is also affected and make the updates there? Should be quick™


ratheron:
I've checked the competition action, the name change shouldnt break anything

ratheron (approved):

PR #86, opened 2026-04-29
Fix close cleanup in real race env

Yuming-Lee24:
## Changes

- Remove early-return guard in [`close()`](lsy_drone_racing/envs/real_race_env.py#L445)
- Guard `_return_to_start()` with `if self.data.taken_off`
- Emergency stop, `close_link()`, and ROS connector close now always run via the `try/finally` chain

### DIFF
--- a/lsy_drone_racing/envs/real_race_env.py
+++ b/lsy_drone_racing/envs/real_race_env.py
@@ -449,11 +449,9 @@ class RealRaceCoreEnv:
         Irrespective of succeeding or not, the drone will be stopped immediately afterwards or in
         case of errors, and close the connections to the ROSConnector.
         """
-        if not self.data.drone_connected or not self.data.taken_off:
-            self._ros_connector.close()
-            return
         try:
-            self._return_to_start()
+            if self.data.taken_off:
+                self._return_to_start()
         finally:
             try:
                 # Kill the drone


ratheron:
Have you also tested this in real?

Yuming-Lee24:
> Have you also tested this in real?

no I didn't. Yes, I should have tested this in real before open the PR.

ratheron:
Is this tested now? Then we can merge it. 
Please remove the lock file from this PR

Yuming-Lee24:
the gap happens for a connected but not taken off drone hit the early return and will be left armed with an radio link and no emergency stop.

what I did to test out:

Build a script that create the env like deploy.py does, and calls env.reset() to connect and arm the drone, but not call env.step() to not take off. By decoding the supervisor.info log to get the drone status (bit 1 = is_armed, bit 3 = can_fly, etc.).

Old env.close(): 
state after reset: ['is_armed', 'can_fly'] (raw=0x020A)
state after close(): ['is_armed', 'can_fly'] (raw=0x020A)

Fix env.close():
state after reset: ['is_armed', 'can_fly'] (raw=0x020A)
state after emergency-stop: ['is_locked'] (raw=0x0240)

The fix disarmed the drone. so the fix is valid.

PR #90, opened 2026-05-10
Fix setup_acados.sh

N0OBSTUDENT:
The former acados setup script OVERWRITEs the whole environment variable LD_LIBRARY_PATH, making other library paths not visible.

### DIFF
--- a/tools/setup_acados.sh
+++ b/tools/setup_acados.sh
@@ -62,7 +62,7 @@ fi
 if [ -f ${ACADOS_DIR}/lib/libacados.so ]; then
   export ACADOS_SOURCE_DIR="$ACADOS_DIR"
   export ACADOS_INSTALL_DIR="$ACADOS_DIR"
-  export LD_LIBRARY_PATH="$ACADOS_DIR/lib"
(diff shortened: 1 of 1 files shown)


tools/setup_acados.sh
```diff
@@ -62,7 +62,7 @@ fi
 if [ -f ${ACADOS_DIR}/lib/libacados.so ]; then
   export ACADOS_SOURCE_DIR="$ACADOS_DIR"
   export ACADOS_INSTALL_DIR="$ACADOS_DIR"
-  export LD_LIBRARY_PATH="$ACADOS_DIR/lib"
+  export LD_LIBRARY_PATH="$ACADOS_DIR/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```
ratheron, line 65: Does this work even if the LD_LIBRARY_PATH is not set?

N0OBSTUDENT: Yes, I suppose this is equivalent to something like
```
LD_LIBRARY_PATH = f"{ACADOS_DIR}/lib" + (f":{LD_LIBRARY_PATH}" if LD_LIBRARY else "")
```
in python

ratheron (approved):

PR #102, opened 2026-06-09
Multi-drone real race support via host-client ROS2 architecture

N0OBSTUDENT:
Addition of a client-side environment for real multi-drone races, along with supporting utilities for robust ROS2-based communication and clock synchronization.


ratheron:
Refer to #74

rducrist:
Block diagram of the current state. 
image

rducrist:
adr
Updated scheme as SVG

pyproject.toml
```diff
@@ -40,7 +40,7 @@ gpu = ["jax[cuda12]"]
 # option[deploy]: deploy to real drones
 deploy = [
     "cflib2 @ git+https://github.com/ratheron/crazyflie-lib-python-v2.git@bccd34092c22392b941bbef9bc5a3718b9d27da0",
-    "drone-estimators",
+    "drone-estimators @ git+https://github.com/learnsyslab/drone-estimators.git",
```
ratheron: I think the better way would be to brush up drone-estimators, release a new version to pypi, and then pin it here

ratheron (commented):
Looking good. I like the clearer naming of the messages. What is still missing for completion?

config/multi_level2.toml
Yuming-Lee24, line 64: why is this control frequency faster then the other one?

Yuming-Lee24:
in order to enable deployment on 2 machines, zenoh router should be runned on both machines and set the listen point. on st-04:
```
 ZENOH_CONFIG_OVERRIDE='listen/endpoints=["tcp/0.0.0.0:7447"]' \
  pixi run -e deploy zenoh-router
```
on st-14:
```
ZENOH_CONFIG_OVERRIDE='connect/endpoints=["tcp/10.157.163.116:7447"]' \
  pixi run -e deploy zenoh-router
```
and mocap, estimators, host, can all runned on of st-14. The latency haven't been quantified.

rducrist:
Verified this. Currently, as we use the default zenoh rmw session config, we set the endpoint to localhost 
```
/// ROS setting: By default connect to the Zenoh router on localhost on port 7447.
    endpoints: [
      "tcp/localhost:7447"
    ], [(https://vscode.dev/github/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5).]
```
We don't need to run a zenoh-router per machine, only one router on the host. The client machines just have to override their endpoints to the IP of the host. 
We could also assert the host a custom IP I guess to make it machine independent. I did a latency check between two workstations in the lab **st11** and **st4**. 
`n=1000 min=0.808 ms avg=1.377 ms median=1.129 ms max=6.492 ms`
Need further experiments to check acutal flight performance.

Yuming-Lee24:
> Verified this. Currently, as we use the default zenoh rmw session config, we set the endpoint to localhost
> 
> ```
> /// ROS setting: By default connect to the Zenoh router on localhost on port 7447.
>     endpoints: [
>       "tcp/localhost:7447"
>     ], [(https://vscode.dev/github/ros2/rmw_zenoh/blob/rolling/rmw_zenoh_cpp/config/DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5).]
> ```
> 
> We don't need to run a zenoh-router per machine, only one router on the host. The client machines just have to override their endpoints to the IP of the host. We could also assert the host a custom IP I guess to make it machine independent. I did a latency check between two workstations in the lab **st11** and **st4**. `n=1000 min=0.808 ms avg=1.377 ms median=1.129 ms max=6.492 ms` Need further experiments to check acutal flight performance.

today I tried to set up according to this on st14 and st 04. 

A known fact is that st-04 cannot establish a TCP connection to the Zenoh router on st-14 at 192.168.1.58:7447; the connection attempts time out even when the router is listening. Connections in the reverse direction—from st-14 to st-04 on port 7447—work.

I runned the router on st-04, with setting the endpoints to listen 7447, and it works. However, it did not work with the same config on st-14 while running router on st-14 because of the fact above.

Yuming-Lee24:
our finding to get one router working:
on st-04:
`pixi shell -e deploy`
`pixi run zenoh-router`
on st-14:
`pixi shell -e deploy`
`export ZENOH_CONFIG_OVERRIDE='mode="client";connec
t/endpoints=["tcp/10.157.163.116:7447"]'`
to check connection, run this on st-14:
`ss -tnp | grep 7447`
further test would be:
st04:
`ros2 run demo_nodes_cpp talker`
st14:
`ros2 run demo_nodes_cpp listener`
message can be published and subscribed
latency check:
`n=3000 min=1.420 ms avg=2.156 ms median=2.151 ms max=2.812 ms`
cc: @rducrist

ratheron:
Just to clarify, st4 is till hooked up to the router, right? This complicates things. We should just get st4 into the main network. Either by letting the router forward both ways (right now we can reach the main network from the routers, but not the other way) or by connecting the st4 directly to the switch

Yuming-Lee24:
> Just to clarify, st4 is till hooked up to the router, right? This complicates things. We should just get st4 into the main network. Either by letting the router forward both ways (right now we can reach the main network from the routers, but not the other way) or by connecting the st4 directly to the switch

I think it is the st14(the ADR PC) hooked with the ASUS router, no?

Yuming-Lee24:
new experiment:
st04: zenoh router, mocap, 2 estimators, client 0(cf01), host
st14: client 1(cf11)
both runs attitude_mpc_multi.py, surprisingly, turns out client 1 performs better then client 0 (less jittering).

Yuming-Lee24:
> new experiment: st04: zenoh router, mocap, 2 estimators, client 0(cf01), host st14: client 1(cf11) both runs attitude_mpc_multi.py, surprisingly, turns out client 1 performs better then client 0 (less jittering).

I should have do it again the other way around, maybe tomorrow, too tired today

ratheron:
> > Just to clarify, st4 is till hooked up to the router, right? This complicates things. We should just get st4 into the main network. Either by letting the router forward both ways (right now we can reach the main network from the routers, but not the other way) or by connecting the st4 directly to the switch
> 
> I think it is the st14(the ADR PC) hooked with the ASUS router, no?

Yes, I meant 14

PR #103, opened 2026-06-11
Upgrade to cflib2

ratheron:
Current cflib 1 implementation is slow. Added

- full crazyflie wrapper in cflib 2
- test script to test the wrapper (state, attitude commands, goto, return function, emergency stop) outside the deploy environments

Tested with the testing script and the regular deployment

### DIFF
--- a/lsy_drone_racing/envs/real_race_env.py
+++ b/lsy_drone_racing/envs/real_race_env.py
@@ -8,35 +8,24 @@ real drones.
 
 from __future__ import annotations
 
-import logging
-import multiprocessing as mp
-import struct
 import time
 from dataclasses import dataclass
-from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
-import cflib
 import jax
 import numpy as np
 import rclpy
-from cflib.crazyflie import Crazyflie, Localization
-from cflib.crtp.crtpstack import CRTPPacket, CRTPPort
-from cflib.utils.power_switch import PowerSwitch
 from drone_estimators.ros_nodes.ros2_connector import ROSConnector
 from drone_models.core import load_params
-from drone_models.transform import force2pwm
 from gymnasium import Env
-from scipy.spatial.transform import Rotation as R
 
 from lsy_drone_racing.envs.utils import gate_passed, load_track
 from lsy_drone_racing.utils.checks import check_drone_start_pos, check_race_track
+from lsy_drone_racing.utils.crazyflie import RacingCrazyflie
 
 if TYPE_CHECKING:
     from ml_collections import ConfigDict
     from numpy.typing import NDArray
-
-logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -89,6 +78,7 @@ class RealRaceCoreEnv:
         randomizations: ConfigDict,
         sensor_range: float = 0.5,
         control_mode: Literal["state", "attitude"] = "state",
+        publish_ros_cmd: bool = True,
     ):
         """Create a deployable version of the drone racing environment.
 
@@ -101,6 +91,7 @@ class RealRaceCoreEnv:
             sensor_range: Sensor range. Determines at which distance the exact position of the
                 gates and obstacles is reveiled.
             control_mode: Control mode of the drone.
+            publish_ros_cmd: Whether to publish attitude commands to ROS for the estimator.
         """
         assert rclpy.ok(), "ROS2 is not running. Please start ROS2 before creating a deploy env."
         # Static env data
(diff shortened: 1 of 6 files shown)


lsy_drone_racing/envs/real_race_env.py
```diff
@@ -8,36 +8,25 @@
...
-from cflib.utils.power_switch import PowerSwitch
 from drone_estimators.ros_nodes.ros2_connector import ROSConnector
 from drone_models.core import load_params
-from drone_models.transform import force2pwm
 from gymnasium import Env
-from scipy.spatial.transform import Rotation as R
 
 from lsy_drone_racing.envs.utils import gate_passed, load_track
 from lsy_drone_racing.utils.checks import check_drone_start_pos, check_race_track
+from lsy_drone_racing.utils.crazyflie import RacingCrazyflie
```
amacati: Why not just `Crazyflie`?

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -162,22 +149,20 @@ def _reset(self, *, seed: int | None = None, options: dict | None = None) -> tup
             )
         self.data.reset(np.stack([self._ros_connector.pos[n] for n in self.drone_names]))
 
-        self._connect_radio(
-            radio_id=self.rank, radio_channel=self.drone_channel, drone_id=self.drone_id
-        )
+        self.drone.connect_and_reset(timeout=10.0, unlock_thrust=self.control_mode == "attitude")
+        self.data.drone_connected = True
```
amacati: Do we still need this after the changes? Seems unused

lsy_drone_racing/utils/crazyflie.py
```diff
@@ -0,0 +1,378 @@
...
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["RacingCrazyflie"]
+
+_DISCONNECT_ERRORS = (DisconnectedError, LinkError, TimeoutError)
+_POWER_CYCLE_BOOT_WAIT = 3.0
+
+
+class RacingCrazyflie:
```
amacati: Mb just `Crazyflie` or `SyncCrazyflie` or `CrazyflieWrapper`, nothing here is `Racing`

lsy_drone_racing/utils/crazyflie.py
```diff
@@ -0,0 +1,378 @@
...
+
+    @property
+    def is_connected(self) -> bool:
+        """Return whether the Crazyflie is currently connected."""
+        return self.active and self.cf is not None
+
+    @property
+    def is_healthy(self) -> bool:
+        """Return whether commands can still be sent to the Crazyflie."""
+        return self.is_connected
```
amacati: Redundant?

lsy_drone_racing/utils/crazyflie.py
```diff
@@ -0,0 +1,378 @@
...
+
+    def connect_and_reset(self, timeout: float = 10.0, unlock_thrust: bool = False) -> None:
+        """Connect, apply race settings, arm, reset the estimator, and optionally unlock thrust."""
+        self._run(self._connect(timeout))
+        self._run(self._apply_settings())
+        self._run(self._reset_estimator())
+        self._run(self._arm())
+        if unlock_thrust:
+            self._run_command("Unlocking thrust", self._unlock_thrust())
```
amacati: Split into `connect()` and `reset()` if you want a cleaner API

lsy_drone_racing/utils/crazyflie.py
```diff
@@ -0,0 +1,378 @@
...
+                self._ros_connector.publish_cmd(command)
+            return
+
+        if control_mode == "state":
+            self._run_command(
+                "Position setpoint", self._send_position_setpoint(action[:3], action[9])
+            )
+            # We don't publish the state command to ros, since the estimator only handles 4D arrays
+            # TODO Maybe we should publish pos+yaw commands, could be interesting for debugging
+            return
```
amacati: Asserting control shapes would be good as a sanity check. Also, are we sure we do not want `send_attitude_control(..., publish=True)` and `send_state_control()`?

lsy_drone_racing/utils/crazyflie.py
```diff
@@ -0,0 +1,378 @@
...
+                    self._mark_disconnected("Emergency stop", exc)
+                except CrazyflieError as exc:
+                    logger.warning(f"Emergency stop failed during shutdown: {exc}")
+
+            if self.cf is not None:
+                self._run(self._disconnect())
+        finally:
+            self._loop.close()
+            self._ros_connector.close()
+            self._closed = True
```
amacati: Could define a 
```python
@property
def closed(self) -> bool
    return self._loop.closed and self._ros_connector.closed
```
that would expose this without manual bookkeeping. Nicer if we want to check this somewhere else

lsy_drone_racing/utils/crazyflie.py
```diff
@@ -0,0 +1,378 @@
...
+        try:
+            self._run(coroutine)
+        except _DISCONNECT_ERRORS as exc:
+            self._mark_disconnected(action_name, exc)
+
+    def _cf(self) -> CflibCrazyflie:
+        if not self.is_connected:
+            raise RuntimeError(f"Drone {self.uri} is not active.")
+        assert self.cf is not None
+        return self.cf
```
amacati: Would make more sense to have a
```python
@property
def cf(self) -> CflibCrazyflie | None:
    return self._cf
```
which is always "safe" and a `self._cf` that might be unsafe. Needs some changes throughout the wrapper in places where you use the current version of `self.cf`.

amacati (commented):
Minor comments, mainly naming stuff and unused params

lsy_drone_racing/utils/crazyflie.py
ratheron: The ROS connector doesnt have a closed field and we don't really need to check for closed. However, we can use the loop to check, if we are allowed to connect. Removed the manual book keeping

lsy_drone_racing/utils/crazyflie.py
ratheron: Agreed

ratheron:
I updated the PR and tested it in real. The current cflib2 implementation does not support full state commands yet. Do we still do the transition already?

ratheron:
Implemented and tested it in my cflib2 fork. Version can later be pinned to official one, once those two PRs are done:
https://github.com/bitcraze/crazyflie-lib-rs/pull/62
https://github.com/bitcraze/crazyflie-lib-python-v2/pull/13
Done from my side

tests/deploy/test_deployment.py
```diff
@@ -0,0 +1,260 @@
...
+
+def main() -> None:
+    """Run the manual deployment smoke test."""
+    import rclpy
+    from drone_estimators.ros_nodes.ros2_connector import ROSConnector
+    from drone_models.core import load_params
+
+    from lsy_drone_racing.control.attitude_controller import AttitudeController
+    from lsy_drone_racing.utils import load_config
+    from lsy_drone_racing.utils.crazyflie import Crazyflie
```
amacati: Import top of the file

amacati (approved):

PR #108, opened 2026-06-25
Switch docs to properdocs (formally mkdocs)

amacati:
Unifies our docs handling to match other projects. Also simplifies local development with the same pixi setup as in crazyflow.


amacati:
Shows a lot of changes, but most of them are renames. Unfortunately, GitHub does not allow control over which files to diff against which. To review this more easily, you can run this script:

```bash
#!/usr/bin/env bash
# Review the rst -> md docs conversion (vs main) as clean per-file diffs:
# main's docs are path-aligned (.rst renamed to .md), so each page reviews as a
# plain .md -> .md content diff with no rename-detection guessing.
#
#   scripts/review_docs_diff.sh                 # everything
#   scripts/review_docs_diff.sh docs/index.md   # a single file
#
# Set DOCS_DIFF_COLOR=never for plain output (e.g. when saving to a file).
set -uo pipefail

TMP="$(mktemp -d)"; trap 'rm -rf "$TMP"' EXIT
git archive main docs | tar -x -C "$TMP"
find "$TMP/docs" -name '*.rst' -exec bash -c 'mv "$1" "${1%.rst}.md"' _ {} \;

COLOR="${DOCS_DIFF_COLOR:-always}"
TARGET="${1:-docs}"   # no arg -> whole docs tree; otherwise the single file given

git --no-pager diff --no-index --color="$COLOR" "$TMP/$TARGET" "$TARGET" | sed "s#$TMP/##g" || true
```

Paste it into the repository and view individual file changes as, e.g., `./review_docs_diff.sh docs/index.md`.

ratheron (approved):

PR #119, opened 2026-07-14
Add warning for controller execution time exceeding loop frequency in sim

Yuming-Lee24:
mirroring deploy.py, the sim will now also shows warning message if the control exceeds loop frequency

### DIFF
--- a/scripts/sim.py
+++ b/scripts/sim.py
@@ -82,18 +82,17 @@ def simulate(
         while True:
             curr_time = i / config.env.freq
 
-            t_start = time.perf_counter()
+            t_loop = time.perf_counter()
             action = controller.compute_control(obs, info)
-            control_time = time.perf_counter() - t_start
+            control_time = time.perf_counter() - t_loop
 
             obs, reward, terminated, truncated, info = env.step(action)
             # Update the controller internal state and models.
-            t_start = time.perf_counter()
+            t_callback = time.perf_counter()
             controller_finished = controller.step_callback(
                 action, obs, reward, terminated, truncated, info
             )
-            control_time += time.perf_counter() - t_start
-            if (exc := control_time - 1 / config.env.freq) > 0:
+            if (exc := time.perf_counter() - t_callback + control_time - 1 / config.env.freq) > 0:
                 logger.warning(f"Controller execution time exceeded loop frequency by {exc:.3f}s.")
             # Add up reward, collisions
             if terminated or truncated or controller_finished:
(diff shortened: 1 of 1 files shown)


ratheron (changes requested):
Why not keep it simpler?

ratheron:
Since multiple teams this term noticed towards the end that their control approach is not feasible in real due to recomputation, I think this PR helps with that. What do you think @amacati

amacati:
Seems very useful, thanks. @ratheron Looking at the review, why not measure `compute_control`? That's usually where the main computation burden is in

scripts/sim.py
```diff
@@ -81,13 +82,19 @@ def simulate(
         while True:
             curr_time = i / config.env.freq
 
+            t_start = time.perf_counter()
```
ratheron:
```suggestion
            t_loop = time.perf_counter()
```

scripts/sim.py
```diff
@@ -81,13 +82,19 @@ def simulate(
...
             action = controller.compute_control(obs, info)
+            control_time = time.perf_counter() - t_start
```
ratheron:
```suggestion
            control_time = time.perf_counter() - t_loop
```

scripts/sim.py
```diff
@@ -81,13 +82,19 @@ def simulate(
...
 
             obs, reward, terminated, truncated, info = env.step(action)
             # Update the controller internal state and models.
+            t_start = time.perf_counter()
```
ratheron:
```suggestion
            t_callback = time.perf_counter()
```

scripts/sim.py
```diff
@@ -81,13 +82,19 @@ def simulate(
...
             controller_finished = controller.step_callback(
                 action, obs, reward, terminated, truncated, info
             )
+            control_time += time.perf_counter() - t_start
+            if (exc := control_time - 1 / config.env.freq) > 0:
+                logger.warning(f"Controller execution time exceeded loop frequency by {exc:.3f}s.")
```
ratheron:
```suggestion
            if (exc := time.perf_counter() - t_callback + control_time - 1 / config.env.freq) > 0:
                logger.warning(f"Controller execution time exceeded loop frequency by {exc:.3f}s.")
```

ratheron (changes requested):

ratheron:
You are right, was unintentional. I adjusted my review

amacati (approved):
Looks good to me, thanks for the addition!

ratheron (approved):
Thanks!

PR #121, opened 2026-07-17
Add gate order field

ratheron:
Added multi lap to ADR.

The gates now have to be passed in `gate_order`, given in the `level.toml`. The sign of the gate number can be negative to indicate reverse direction. Changed the `gate_passed` function and added a `load_gate_order` helper. Adapted the envs (mainly the obs) and docs accordingly.

### DIFF
--- a/lsy_drone_racing/envs/race_core.py
+++ b/lsy_drone_racing/envs/race_core.py
@@ -74,7 +74,8 @@ class EnvData:
     episode.
 
     Attributes:
-        race_progress: Current target waypoint index for each drone in each environment
+        n_gates_passed: Number of configured gate-order entries already passed by each drone in
+            each environment
         gates_visited: Boolean flags indicating which gates have been visited by each drone
         obstacles_visited: Boolean flags indicating which obstacles have been detected
         last_drone_pos: Previous positions of drones, used for gate passing detection
@@ -87,12 +88,13 @@ class EnvData:
         obstacle_mj_ids: MuJoCo IDs for the obstacles
         max_episode_steps: Maximum number of steps per episode
         sensor_range: Range at which drones can detect gates and obstacles
-        gate_order_ids: 0-based gate IDs describing the configured gate order
-        gate_order_reverse: Flags indicating whether each gate-order entry is reversed
+        gate_sequence: 0-based gate IDs describing the configured gate order
+        gate_sequence_direction: Signed passing directions for each gate-order entry. ``1``
+            indicates the positive passing direction, ``-1`` the reverse direction.
     """
 
     # Dynamic variables
-    race_progress: Array
+    n_gates_passed: Array
     gates_visited: Array
     obstacles_visited: Array
     last_drone_pos: Array
@@ -109,8 +111,8 @@ class EnvData:
     nominal_gates_pos: Array
     nominal_gates_quat: Array
     nominal_obstacles_pos: Array
-    gate_order_ids: Array
-    gate_order_reverse: Array
+    gate_sequence: Array
+    gate_sequence_direction: Array
     # sim_data is stored in the env data to allow passing a single tree on which we can operate
     sim_data: SimData
     # Static variables
@@ -132,8 +134,8 @@ class EnvData:
         nominal_gates_pos: Array,
         nominal_gates_quat: Array,
         nominal_obstacles_pos: Array,
-        gate_order_ids: Array,
-        gate_order_reverse: Array,
+        gate_sequence: Array,
+        gate_sequence_direction: Array,
         sim_data: SimData,
         device: Device,
     ) -> EnvData:
@@ -150,7 +152,7 @@ class EnvData:
             jp.tile(nominal_obstacles_pos[None, ...], (n_envs, 1, 1)), device
         )
         return EnvData(
-            race_progress=jp.zeros((n_envs, n_drones), dtype=int, device=device),
+            n_gates_passed=jp.zeros((n_envs, n_drones), dtype=int, device=device),
             gates_visited=jp.zeros((n_envs, n_drones, n_gates), dtype=bool, device=device),
             obstacles_visited=jp.zeros((n_envs, n_drones, n_obstacles), dtype=bool, device=device),
             last_drone_pos=jp.zeros((n_envs, n_drones, 3), dtype=np.float32, device=device),
@@ -168,8 +170,8 @@ class EnvData:
             nominal_gates_pos=tiled_gates_pos,
             nominal_gates_quat=tiled_gates_quat,
             nominal_obstacles_pos=tiled_obstacles_pos,
-            gate_order_ids=jp.array(gate_order_ids, dtype=int, device=device),
-            gate_order_reverse=jp.array(gate_order_reverse, dtype=bool, device=device),
+            gate_sequence=jp.array(gate_sequence, dtype=int, device=device),
+            gate_sequence_direction=jp.array(gate_sequence_direction, dtype=int, device=device),
             sim_data=sim_data,
             sensor_range=jp.array([sensor_range], dtype=jp.float32, device=device),
         )
@@ -257,9 +259,11 @@ def build_observation_space(n_gates: int, n_obstacles: int, n_gate_passes: int) 
         "quat": spaces.Box(low=-1, high=1, shape=(4,)),
         "vel": spaces.Box(low=-np.inf, high=np.inf, shape=(3,)),
         "ang_vel": spaces.Box(low=-np.inf, high=np.inf, shape=(3,)),
-        "race_progress": spaces.Discrete(n_gate_passes + 1, start=-1),
-        "target_gate": spaces.Discrete(n_gates + 1, start=-1),
-        "target_gate_reverse": spaces.Box(low=0, high=1, shape=(), dtype=bool),
+        "n_gates_passed": spaces.Discrete(n_gate_passes + 1, start=0),
+        "gate_sequence": spaces.MultiDiscrete(np.full(n_gate_passes, n_gates, dtype=np.int64)),
+        "gate_sequence_direction": spaces.Box(
+            low=-1, high=1, shape=(n_gate_passes,), dtype=np.int32
+        ),
         "gates_pos": spaces.Box(low=-np.inf, high=np.inf, shape=(n_gates, 3)),
         "gates_quat": spaces.Box(low=-1, high=1, shape=(n_gates, 4)),
         "gates_visited": spaces.Box(low=0, high=1, shape=(n_gates,), dtype=bool),
@@ -297,8 +301,11 @@ class RaceCoreEnv:
     * quat: Drone orientation as a quaternion (x, y, z, w)
     * vel: Drone linear velocity
     * ang_vel: Drone angular velocity
-    * race_progress: The current index in the configured gate order, or -1 if the track is
-      completed
+    * n_gates_passed: The number of entries in the configured gate order that have already been
+      passed
+    * gate_sequence: The configured gate order expressed as 0-based gate IDs
+    * gate_sequence_direction: Signed passing direction for each gate-sequence entry. `1`
+      indicates the positive passing direction, `-1` the reverse direction
     * gates_pos: Positions of the gates
     * gates_quat: Orientations of the gates
     * gates_visited: Flags indicating if the drone already was/ is in the sensor range of the
@@ -306,9 +313,6 @@ class RaceCoreEnv:
     * obstacles_pos: Positions of the obstacles
     * obstacles_visited: Flags indicating if the drone already was/ is in the sensor range of the
       obstacles and the true position is known
-    * target_gate: The 0-based gate ID of the next gate in the configured gate order, or -1 if
-      the track is completed
-    * target_gate_reverse: Whether the next gate must be passed in reverse direction
 
     The action space consists of a desired full-state command
     [x, y, z, vx, vy, vz, ax, ay, az, yaw, rrate, prate, yrate] that is tracked by the drone's
@@ -387,7 +391,7 @@ class RaceCoreEnv:
         self.track = track
         gates, obstacles, drones = load_track(track)
         n_gates, n_obstacles = len(track.gates), len(track.obstacles)
-        gate_order_ids, gate_order_reverse = load_gate_order(track, n_gates)
+        gate_sequence, gate_sequence_direction = load_gate_order(track, n_gates)
         contact_masks = _load_contact_masks(self.sim)
         specs = {} if disturbances is None else disturbances
         disturbances = {mode: rng_spec2fn(spec) for mode, spec in specs.items()}
@@ -416,8 +420,8 @@ class RaceCoreEnv:
             nominal_gates_pos=gates.nominal_pos,
             nominal_gates_quat=gates.nominal_quat,
             nominal_obstacles_pos=obstacles.nominal_pos,
-            gate_order_ids=gate_order_ids,
-            gate_order_reverse=gate_order_reverse,
+            gate_sequence=gate_sequence,
+            gate_sequence_direction=gate_sequence_direction,
             sim_data=self.sim.data,
             device=self.settings.device,
         )
@@ -694,10 +698,6 @@ class RaceCoreEnv:
 
 def obs(data: EnvData) -> dict[str, Array]:
     """Return the observation of the environment."""
-    target_gate = data.gate_order_ids[data.race_progress]
-    target_gate = jp.where(data.race_progress < 0, -1, target_gate)
-    target_gate_reverse = data.gate_order_reverse[data.race_progress]
-    target_gate_reverse = jp.where(data.race_progress < 0, False, target_gate_reverse)
     mask = data.gates_visited[..., None]
     sensor_gates_pos = jp.where(mask, data.gates_pos[:, None], data.nominal_gates_pos[:, None])
     sensor_gates_quat = jp.where(mask, data.gates_quat[:, None], data.nominal_gates_quat[:, None])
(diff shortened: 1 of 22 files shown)


lsy_drone_racing/envs/utils.py
```diff
@@ -86,7 +119,11 @@ def gate_passed(
     pos_local = gate_rot.apply(drone_pos - gate_pos, inverse=True)
     # Check the plane intersection. If passed, calculate the point of the intersection and check if
     # it is within the gate box.
-    passed_plane = (last_pos_local[0] < 0) & (pos_local[0] > 0)
+    passed_plane = jp.where(
+        reverse,
+        (last_pos_local[0] > 0) & (pos_local[0] < 0),
+        (last_pos_local[0] < 0) & (pos_local[0] > 0),
+    )
```
amacati: Easier to read
```suggestion
    passed_plane_fwd = (last_pos_local[0] < 0) & (pos_local[0] > 0)
    passed_plane_rev = (last_pos_local[0] > 0) & (pos_local[0] < 0)
    passed_plane = jp.where(reverse, passed_plane_rev, passed_plane_fwd)
```

lsy_drone_racing/envs/utils.py
```diff
@@ -51,23 +52,54 @@ def load_track(track: ConfigDict) -> tuple[ConfigDict, ConfigDict, ConfigDict]:
...
+    gate_order = np.asarray(track["gate_order"], dtype=np.int32)
+    assert gate_order.ndim == 1, "track.gate_order must be a 1D list of signed integers."
+    assert gate_order.size > 0, "track.gate_order must be a non-empty list."
+    assert not np.any(gate_order == 0), "track.gate_order must not contain 0. "
+    gate_sequence = np.abs(gate_order) - 1
+    if np.any(gate_sequence < 0) or np.any(gate_sequence >= n_gates):
+        raise ValueError(
+            f"track.gate_order must reference gate numbers in the interval [1, {n_gates})."
+        )
+    gate_sequence_direction = np.where(gate_order < 0, -1, 1).astype(np.int32)
```
amacati: gate_order does not have 0, so this should be fine

```suggestion
    gate_sequence_direction = np.sign(gate_order)
```

lsy_drone_racing/envs/race_core.py
```diff
@@ -768,17 +795,20 @@ def _update_visited_objects(data: EnvData) -> EnvData:
 
 
 def _update_target_gates(data: EnvData) -> EnvData:
-    """Update the target gate index based on the current target gate and the number of gates."""
-    n_gates = data.gates_pos.shape[1]
+    """Update the waypoint progress based on the current gate-order entry."""
```
amacati: Why waypoint? Should be number of passed gates or something

lsy_drone_racing/envs/race_core.py
```diff
@@ -768,17 +795,20 @@ def _update_visited_objects(data: EnvData) -> EnvData:
...
     gates_pos, gates_quat = data.gates_pos, data.gates_quat
     drone_pos = data.sim_data.states.pos
-    gate_pos = gates_pos[jp.arange(gates_pos.shape[0])[:, None], data.target_gate % n_gates]
-    gate_quat = gates_quat[jp.arange(gates_quat.shape[0])[:, None], data.target_gate % n_gates]
-    passed = gate_passed(drone_pos, data.last_drone_pos, gate_pos, gate_quat, (0.45, 0.45))
-    # Update the target gate index. Increment by one if drones have passed a gate
-    target_gate = data.target_gate + passed * ~data.disabled_drones
-    target_gate = jp.where(target_gate >= n_gates, -1, target_gate)
-    return data.replace(target_gate=target_gate, last_drone_pos=data.sim_data.states.pos)
+    progress = jp.clip(data.n_gates_passed, 0, n_gate_passes - 1)
```
amacati: Why the clipping? This is guaranteed, is it not? Also, I'm not sure what is meant by progress. It is still `n_gate_passes`, right?

lsy_drone_racing/envs/race_core.py
```diff
@@ -768,17 +795,20 @@ def _update_visited_objects(data: EnvData) -> EnvData:
...
+    gate_ids = data.gate_sequence[progress]
+    reverse = data.gate_sequence_direction[progress] < 0
+    gate_pos = gates_pos[jp.arange(gates_pos.shape[0])[:, None], gate_ids]
+    gate_quat = gates_quat[jp.arange(gates_quat.shape[0])[:, None], gate_ids]
+    passed = gate_passed(drone_pos, data.last_drone_pos, gate_pos, gate_quat, reverse, (0.45, 0.45))
+    # Advance the gate-order progress by one if drones have passed the current waypoint.
+    increment = (passed & ~data.disabled_drones).astype(data.n_gates_passed.dtype)
+    n_gates_passed = jp.minimum(data.n_gates_passed + increment, n_gate_passes)
```
amacati: `increment` already prevents adding beyond n_gates_passed, right? So we don't need the subsequent jp.minimum.

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -174,15 +175,19 @@ def _step(self, action: NDArray) -> tuple[dict, float, bool, bool, dict]:
         dpos = drone_pos[:, None, :2] - self.obstacles.pos[None, :, :2]
         self.data.obstacles_visited |= np.linalg.norm(dpos, axis=-1) < self.sensor_range
 
-        gate_pos = self.gates.pos[self.data.target_gate]
-        gate_quat = self.gates.quat[self.data.target_gate]
+        n_gate_passes = self.gate_sequence.shape[0]
+        progress = np.clip(self.data.n_gates_passed, 0, n_gate_passes - 1)
```
amacati: Why rename it, and why do we clip it? It should not be possible to be anything else, and if it is, we would like to know by throwing an error, right?

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -174,15 +175,19 @@ def _step(self, action: NDArray) -> tuple[dict, float, bool, bool, dict]:
...
 
         with jax.default_device(self.device):  # Ensure gate_passed runs on the CPU
             passed = gate_passed(
-                drone_pos, self.data.last_drone_pos, gate_pos, gate_quat, (0.45, 0.45)
+                drone_pos, self.data.last_drone_pos, gate_pos, gate_quat, gate_reverse, (0.45, 0.45)
             )
-        self.data.target_gate += np.asarray(passed)
-        self.data.target_gate[self.data.target_gate >= self.n_gates] = -1
+        increment = np.asarray(passed, dtype=self.data.n_gates_passed.dtype)
+        self.data.n_gates_passed = np.minimum(self.data.n_gates_passed + increment, n_gate_passes)
```
amacati: Same here. Do we really need this clamping?

scripts/multi_sim.py
```diff
@@ -120,7 +118,9 @@ def simulate(
 
             obs, reward, terminated, truncated, info = env.step(actions)
 
-            newly_finished = (obs["target_gate"] == -1) & np.isnan(finish_times)
+            newly_finished = (obs["n_gates_passed"] == obs["gate_sequence"].shape[-1]) & np.isnan(
+                finish_times
+            )
```
amacati: I generally prefer something like this to avoid ugly line wrappings, but it's just style, so feel free to ignore
```suggestion
            track_len = obs["gate_sequence"].shape[-1]
            newly_finished = (obs["n_gates_passed"] == track_len) & np.isnan(finish_times)
```

scripts/sim.py
```diff
@@ -104,23 +102,23 @@ def simulate(
...
-    gates_passed = obs["target_gate"]
-    if gates_passed == -1:  # The drone has passed the final gate
-        gates_passed = len(config.env.track.gates)
-    finished = gates_passed == len(config.env.track.gates)
+    finished = obs["n_gates_passed"] == obs["gate_sequence"].shape[0]
     logger.info(
-        f"Flight time (s): {curr_time}\nFinished: {finished}\nGates passed: {gates_passed}\n"
+        f"Flight time (s): {curr_time}\n"
+        + f"Finished: {finished}\n"
+        + f"Gates passed: {obs['n_gates_passed']}\n"
```
amacati: Should work without
```suggestion
        f"Flight time (s): {curr_time}\n"
        f"Finished: {finished}\n"
        f"Gates passed: {obs['n_gates_passed']}\n"
```

tests/unit/utils/test_utils.py
```diff
@@ -1,12 +1,15 @@
+import os
 from pathlib import Path
 
+os.environ["SCIPY_ARRAY_API"] = "1"
+
```
amacati: Can we set this once in conftest.py instead?

tests/unit/utils/test_utils.py
```diff
@@ -30,23 +33,48 @@ def test_gate_pass():
...
+@pytest.mark.unit
+def test_load_gate_order():
+    config = load_config(Path(__file__).parents[3] / "config/level0.toml")
+    config.env.track.gate_order = [1, -2, 2, -1]
+    gate_sequence, gate_sequence_direction = load_gate_order(
+        config.env.track, len(config.env.track.gates)
+    )
+    np.testing.assert_array_equal(gate_sequence, np.array([0, 1, 1, 0]))
+    np.testing.assert_array_equal(gate_sequence_direction, np.array([1, -1, 1, -1]))
```
amacati, line 77: Can we test that it throws for false input if that is something we want to enforce?

tests/unit/envs/test_race_core.py
```diff
@@ -75,11 +77,13 @@ def test_obs_structure_and_initial_values():
...
     # Single-drone make() squeezes leading (world, drone) dims: pos is (3,), gates_pos is
-    # (n_gates, 3), target_gate is a 0-d scalar.
+    # (n_gates, 3).
     assert np.asarray(obs["pos"]).shape == (3,)
     assert np.asarray(obs["gates_pos"]).ndim == 2
     assert np.asarray(obs["gates_pos"]).shape[1] == 3
-    assert int(np.asarray(obs["target_gate"]).item()) == 0
+    assert int(np.asarray(obs["n_gates_passed"]).item()) == 0
+    np.testing.assert_array_equal(np.asarray(obs["gate_sequence"]), np.array([0, 1, 2, 3]))
+    np.testing.assert_array_equal(np.asarray(obs["gate_sequence_direction"]), np.ones(4))
```
amacati: Can we keep the same logic as before? We just want to assert the shape, not the values

tests/unit/envs/test_race_core.py
```diff
@@ -268,36 +274,78 @@ def test_gate_pass_increments_target_gate():
...
+@pytest.mark.unit
+def test_gate_order_initial_obs_mapping():
+    """Initial observations must expose the configured gate sequence and directions."""
+    config = load_config(CONFIG_PATH / "level0.toml")
+    config.env.track.gate_order = [-2, 1]
+    env = make_env(track=config.env.track)
+    env_obs, _ = env.reset()
+    assert int(np.asarray(env_obs["n_gates_passed"]).item()) == 0
+    np.testing.assert_array_equal(np.asarray(env_obs["gate_sequence"]), np.array([1, 0]))
+    np.testing.assert_array_equal(np.asarray(env_obs["gate_sequence_direction"]), np.array([-1, 1]))
```
amacati: Why are we using this now? Also, if we start using it, we should use `strict=True` and import it as `from numpy.testing import assert_array_equal` to make things clearer.

tests/unit/envs/test_race_core.py
```diff
@@ -268,36 +274,78 @@ def test_gate_pass_increments_target_gate():
...
+    new_last = data.last_drone_pos.at[0, 0, :].set(jp.asarray(behind))
+    new_pos = data.sim_data.states.pos.at[0, 0, :].set(jp.asarray(front))
+    new_sim_data = data.sim_data.replace(states=data.sim_data.states.replace(pos=new_pos))
+    env.unwrapped.data = data.replace(last_drone_pos=new_last, sim_data=new_sim_data)
+
+    new_data = _update_target_gates(env.unwrapped.data)
+    mapped_obs = obs(new_data)
+    assert int(np.asarray(mapped_obs["n_gates_passed"][0, 0])) == 1
+    assert int(np.asarray(mapped_obs["gate_sequence"][0, 0, 1])) == 0
+    assert int(np.asarray(mapped_obs["gate_sequence_direction"][0, 0, 1])) == -1
```
amacati:
```suggestion
    next_data = _update_target_gates(env.unwrapped.data)
    next_obs = obs(next_data)
    assert int(np.asarray(next_obs["n_gates_passed"][0, 0])) == 1
    assert int(np.asarray(next_obs["gate_sequence"][0, 0, 1])) == 0
    assert int(np.asarray(next_obs["gate_sequence_direction"][0, 0, 1])) == -1
```

tests/unit/envs/test_race_core.py
```diff
@@ -369,28 +417,29 @@ def test_gate_not_passed_when_outside_gate_box():
...
+    assert int(np.asarray(new_data.n_gates_passed[0, 0])) == 0
     env.close()
 
 
 @pytest.mark.unit
-def test_gate_pass_at_last_gate_clamps_to_negative_one():
-    """Passing the final gate must set target_gate to -1 (course finished sentinel)."""
+def test_gate_pass_at_last_gate_sets_n_gates_passed_to_sequence_length():
+    """Passing the final gate must set n_gates_passed to the full sequence length."""
     from scipy.spatial.transform import Rotation as R
```
amacati: Why the inline import? We should have no inline imports at all. This has slipped through, can you please fix it while you are at it?

tests/unit/envs/test_race_core.py
```diff
@@ -402,7 +451,37 @@ def test_gate_pass_at_last_gate_clamps_to_negative_one():
     env.unwrapped.data = data.replace(last_drone_pos=new_last, sim_data=new_sim_data)
 
     new_data = _update_target_gates(env.unwrapped.data)
-    assert int(np.asarray(new_data.target_gate[0, 0])) == -1
+    assert int(np.asarray(new_data.n_gates_passed[0, 0])) == n_gate_passes
+    mapped_obs = obs(new_data)
+    assert int(np.asarray(mapped_obs["n_gates_passed"][0, 0])) == n_gate_passes
```
amacati:
```suggestion
    next_obs = obs(new_data)
    assert int(np.asarray(next_obs["n_gates_passed"][0, 0])) == n_gate_passes
```

tests/unit/envs/test_race_core.py
```diff
@@ -402,7 +451,37 @@ def test_gate_pass_at_last_gate_clamps_to_negative_one():
...
+
+    new_last = data.last_drone_pos.at[0, 0, :].set(jp.asarray(front))
+    new_pos = data.sim_data.states.pos.at[0, 0, :].set(jp.asarray(behind))
+    new_sim_data = data.sim_data.replace(states=data.sim_data.states.replace(pos=new_pos))
+    env.unwrapped.data = data.replace(last_drone_pos=new_last, sim_data=new_sim_data)
+
+    new_data = _update_target_gates(env.unwrapped.data)
+    mapped_obs = obs(new_data)
+    assert int(np.asarray(new_data.n_gates_passed[0, 0])) == 1
+    assert int(np.asarray(mapped_obs["n_gates_passed"][0, 0])) == 1
```
amacati:
```suggestion
    next_data = _update_target_gates(env.unwrapped.data)
    next_obs = obs(next_data)
    assert int(np.asarray(next_data.n_gates_passed[0, 0])) == 1
    assert int(np.asarray(next_obs["n_gates_passed"][0, 0])) == 1
```

amacati (changes requested):
Mostly small nits

lsy_drone_racing/envs/real_race_env.py
ratheron: The clamping is there so it doesnt go higher than the max gate order. But I see the point. When we pass the last gate, we take over anyway, so we can never increment higher

tests/unit/envs/test_race_core.py
ratheron: This goes beyond just checking for the shape. Why should we not at least once check for the correct values of the obs? 
Simplified the import

amacati (approved):

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -174,15 +175,16 @@ def _step(self, action: NDArray) -> tuple[dict, float, bool, bool, dict]:
...
+        gate_quat = self.gates.quat[gate_ids]
 
         with jax.default_device(self.device):  # Ensure gate_passed runs on the CPU
             passed = gate_passed(
-                drone_pos, self.data.last_drone_pos, gate_pos, gate_quat, (0.45, 0.45)
+                drone_pos, self.data.last_drone_pos, gate_pos, gate_quat, gate_reverse, (0.45, 0.45)
             )
-        self.data.target_gate += np.asarray(passed)
-        self.data.target_gate[self.data.target_gate >= self.n_gates] = -1
+        self.data.n_gates_passed = self.data.n_gates_passed + passed
```
Yuming-Lee24, line 187:
```suggestion
        self.data.n_gates_passed = self.data.n_gates_passed + np.asarray(passed) 
```
otherwise a TypeError will show up: `TypeError: JAX arrays are immutable and do not support in-place item assignment. Instead of x[idx] = y, use x = x.at[idx].set(y) or another .at[] method: https://docs.jax.dev/en/latest/_autosummary/jax.numpy.ndarray.at.html` . I find this when I am deploying level2 state controller.

amacati: Can you open a PR that fixes this line?

PR #122, opened 2026-07-20
Synthesize noisy nominal track from mocap poses

Yuming-Lee24:
When `env.track.randomize` is set, the real race env now measures the track via mocap at reset and gives the controller a noisy nominal prior (measured − noise sampled from the randomization specs, revealed within sensor range), so level 3 can be deployed directly without running `save_track_as_config.py`.
 
Not sure:
1. Whether to print the noised nominal pos or not.

Test in real: 

Set up a level2 track and use the state controler on level3.toml

1. the `self.gates.nominal_pos` is randomized.
2. the  `self.gates.pos` stays the same as in real and is revealed within the range.
3. the task is finished and auto return is triggered.

A bug I encountered is that when deploying state controller, an Error will raise due to the upgrade of cflib2. 
`TypeError: Commander.send_setpoint_full_state() takes 5 positional arguments but 7 were given`
I will open another PR for this fix.

### DIFF
--- a/lsy_drone_racing/envs/real_race_env.py
+++ b/lsy_drone_racing/envs/real_race_env.py
@@ -8,6 +8,7 @@ to the real drones.
 
 from __future__ import annotations
 
+import logging
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal
 
@@ -17,6 +18,7 @@ import rclpy
 from drone_estimators.ros_nodes.ros2_connector import ROSConnector
 from drone_models.core import load_params
 from gymnasium import Env
+from scipy.spatial.transform import Rotation as R
 
 from lsy_drone_racing.envs.utils import gate_passed, load_gate_order, load_track
 from lsy_drone_racing.utils.checks import check_drone_start_pos, check_race_track
@@ -25,6 +27,8 @@ from lsy_drone_racing.utils.crazyflie import Crazyflie
 if TYPE_CHECKING:
     from ml_collections import ConfigDict
     from numpy.typing import NDArray
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -90,6 +94,7 @@ class RealRaceCoreEnv:
         # Static env data
         self.n_drones = len(drones)
         self.gates, self.obstacles, self.drones = load_track(track)
+        self.randomize_track = track.randomize
         self.n_gates = len(self.gates.pos)
         self.gate_sequence, self.gate_sequence_direction = load_gate_order(track, self.n_gates)
         self.n_obstacles = len(self.obstacles.pos)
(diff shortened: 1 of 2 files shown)


lsy_drone_racing/envs/real_race_env.py
```diff
@@ -90,6 +94,7 @@ def __init__(
         # Static env data
         self.n_drones = len(drones)
         self.gates, self.obstacles, self.drones = load_track(track)
+        self.randomize_track = bool(track.get("randomize", False))
```
amacati: Please program less defensively. We want things to break if they are wrong. randomize should always be defined -> no get, and it should always be a bool -> no casting to bool

```suggestion
        self.randomize_track = track.randomize
```

lsy_drone_racing/envs/real_race_env.py
```diff
@@ -275,6 +287,31 @@ def _update_track_poses(self):
...
+        kwargs = self.randomizations.gate_rpy.kwargs
+        gates_rpy = R.from_quat(self.gates.quat).as_euler("xyz")
+        gates_rpy -= rng.uniform(kwargs.minval, kwargs.maxval, gates_rpy.shape)
+        self.gates.nominal_quat[...] = R.from_euler("xyz", gates_rpy).as_quat()
+        kwargs = self.randomizations.obstacle_pos.kwargs
+        self.obstacles.nominal_pos[...] = self.obstacles.pos - rng.uniform(
+            kwargs.minval, kwargs.maxval, self.obstacles.pos.shape
+        )
+        for i, name in enumerate(self.drone_names):
+            self.drones.pos[i, ...] = self._ros_connector.pos[name]
```
amacati, line 308: Is this required for anything? I think we can just leave it out

amacati (commented):
I like the change, thanks for drafting this! Let me know if it also works with the proposed changes, then we can already merge.

Yuming-Lee24:
> I like the change, thanks for drafting this! Let me know if it also works with the proposed changes, then we can already merge.

Thanks for the review! I will test the changes asap.

lsy_drone_racing/envs/real_race_env.py
Yuming-Lee24: > Is this required for anything? I think we can just leave it out

this block save the take off position in memory, also considered for multi drones. Should I make it only for single drone or leave it out? in that case, user needs to put the drone at the postion that is defined in the toml. I think the current approach is more flexible if I understand it correctly.

amacati: Ah, I see. Students should be able to put down the drone anywhere and launch. I think that's fine, yes

amacati (approved):
Let's merge this. Thanks!

PR #124, opened 2026-07-23
Add pixi tasks for mocap and estimator in deploy env

Yuming-Lee24:
A task on Todo list, add the pixi task to the pyproject to launch mocap and estimator, tested in real --- a drone successfully took off.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,6 +113,12 @@ ros2-distro-mutex = "0.12.*"  # Pinned for motion_capture_tracking
 scripts = ["tools/setup_mocap.sh"]
 
 ## feature [deploy]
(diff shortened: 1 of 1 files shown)


amacati (approved):
Nice! One small change: Does it even make sense to have a default drone on the estimator args? I suspect this will change regularly

Yuming-Lee24:
> Nice! One small change: Does it even make sense to have a default drone on the estimator args? I suspect this will change regularly

yes, true, I was thinking to have a default in pixi task just like having cf1 as default using cli line, but yes, we always want student input this arg. Without a default the pixi will raise an error, which is what we want.

amacati (approved):
Looks good, thanks!

PR #125, opened 2026-07-31
Add live MuJoCo viewer for deployment

Yuming-Lee24:
when using a `--render` or `-r` arg, a viewer window will open when the deployment starts. Tested in real with level3 state controller and attitude controller. One thing need to notice is that `ctrl+C` can not do emergency stop when focusing on the viewer window.

### DIFF
--- a/lsy_drone_racing/utils/deploy_viewer.py
+++ b/lsy_drone_racing/utils/deploy_viewer.py
@@ -1,0 +1,71 @@
+"""Live visualization for real-world deployments."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+
+    from lsy_drone_racing.envs.race_core import RaceCoreEnv
+
+
+class DeployViewer:
+    """Mirror the measured race state in a single-world CPU simulation."""
+
+    def __init__(self, env: RaceCoreEnv):
+        """Take ownership of a render-only core environment.
+
+        Args:
+            env: Single-world, single-drone CPU environment used for rendering.
+        """
+        self._env = env
+
+    def warm_up(self):
+        """Compile rendering and open the window before the real drone is armed."""
+        data = self._env.data
+        sim_data = self._env.sim.data.replace(
+            core=self._env.sim.data.core.replace(
+                mjx_synced=self._env.sim.data.core.mjx_synced.at[...].set(False)
+            )
+        )
+        self._env.data = data.replace(sim_data=sim_data)
+        self._env.render()
+
+    def set_track(self, gates_pos: NDArray, gates_quat: NDArray, obstacles_pos: NDArray):
+        """Update the simulation with the measured track poses.
+
+        Args:
+            gates_pos: Gate positions of shape (n_gates, 3).
+            gates_quat: Gate quaternions in SciPy xyzw order with shape (n_gates, 4).
+            obstacles_pos: Obstacle positions of shape (n_obstacles, 3).
+        """
+        data = self._env.data
+        self._env.data = data.replace(
+            gates_pos=data.gates_pos.at[0].set(gates_pos),
+            gates_quat=data.gates_quat.at[0].set(gates_quat),
+            obstacles_pos=data.obstacles_pos.at[0].set(obstacles_pos),
+        )
+
+    def update(self, pos: NDArray, quat: NDArray):
+        """Update the measured drone pose and render it.
+
+        Args:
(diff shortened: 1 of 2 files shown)


lsy_drone_racing/utils/deploy_viewer.py
```diff
@@ -0,0 +1,124 @@
...
+    for i in range(n_gates):
+        gate = frame.attach_body(gate_body, "", f":{i}")
+        gate.mocap = True
+    obstacle_spec = mujoco.MjSpec.from_file(str(OBSTACLE_SPEC_PATH))
+    if (obstacle_body := obstacle_spec.body("obstacle")) is None:
+        raise ValueError("Obstacle body not found in obstacle spec")
+    for i in range(n_obstacles):
+        obstacle = frame.attach_body(obstacle_body, "", f":{i}")
+        obstacle.mocap = True
+    return spec.compile()
```
amacati: Can we somehow reuse the stuff we are using in our core sim env? We are creating the same spec here, right?

lsy_drone_racing/utils/deploy_viewer.py
```diff
@@ -0,0 +1,124 @@
...
+                )
+            )
+            with handle.lock():
+                camera = handle.cam
+                camera.distance = camera_distance
+                camera.azimuth = camera_azimuth
+                camera.elevation = camera_elevation
+                camera.lookat[:] = camera_lookat
+            self._handle = handle
+            viewer_stack.pop_all()
```
amacati: I wonder, can we just create a 1x1 CPU core env and take the viewer from there? And then write a super thin wrapper around it?

scripts/deploy.py
```diff
@@ -76,7 +95,11 @@ def main(config: str = "level2.toml", controller: str | None = None):
         finished_track = next_obs["n_gates_passed"] == next_obs["gate_sequence"].shape[0]
         logger.info(f"Track time: {ep_time:.3f}s" if finished_track else "Task not completed")
     finally:
-        env.close()
+        try:
+            if viewer is not None:
+                viewer.close()
+        finally:
+            env.close()
```
amacati: Why would the `viewer.close()` fail? Can't we do this?

```suggestion
        env.close()
        if viewer is not None:
            viewer.close()
```

amacati (commented):
Generally a good idea, I just wonder if we can massively simplify things

scripts/deploy.py
Yuming-Lee24: I think would be better if we first close the viewer? Otherwise the viewer will stay frozen untill the drone lands, nothing harmful but feels unnecessary to keep the viewer after the loop finishes.

Yuming-Lee24:
New version is pushed, tested on both level2 and level3.

scripts/deploy.py
```diff
@@ -62,6 +92,8 @@ def main(config: str = "level2.toml", controller: str | None = None):
             obs = {k: v[0] for k, v in obs.items()}
             action = controller.compute_control(obs, info)
             next_obs, reward, terminated, truncated, info = env.step(action)
+            if viewer is not None:
```
amacati:
```suggestion
            if render:
```

scripts/deploy.py
```diff
@@ -47,10 +50,37 @@ def main(config: str = "level2.toml", controller: str | None = None):
...
+                seed=config.env.seed,
+                device="cpu",
+            )
+            viewer = DeployViewer(viewer_env)
+            viewer.warm_up()
+
         obs, info = env.reset(seed=config.env.seed, options=config.deploy)
         next_obs = obs  # Set next_obs to avoid errors when the loop never enters
 
+        if viewer is not None:
```
amacati:
```suggestion
        if render:
```

amacati (commented):
This looks very good and is pretty clean. Just two minor things for readability, they are functionally equal to what you had. Thanks for working on this!

amacati:
Can you pull and launch this one as final sanity check? Then we are ready to merge it @Yuming-Lee24

scripts/deploy.py
amacati: Fair point

Yuming-Lee24:
> Can you pull and launch this one as final sanity check? Then we are ready to merge it @Yuming-Lee24

yes, I will test it tomorrow morning and also check the crazyflow upgrade

amacati:
Perfect, thanks!

Yuming-Lee24:
do not merge before the param loader is fixed

amacati:
That should go in a separate PR

amacati (approved):

PR #126, opened 2026-08-05
Upgrade to crazyflow 0.2.1

amacati:
Crazyflow 0.2 introduced many changes, some of them breaking. The drone models are now directly available in crazyflow, we have named pipelines with targeted insertions/deletions, and the naming scheme has been cleaned up. The changes seem like a lot, but it is mainly renaming and adopting the new API. In places, this is not the most ergonomic, but we will have to wait until https://github.com/learnsyslab/crazyflow/issues/79 gets implemented.


rducrist (approved):
Looking good for me.

PR #127, opened 2026-08-06
Fix parameter loading

Yuming-Lee24:
previously, we use a dynamic param loader to load the drone params (e.g. mass, thrust), additionally specify a dynamic function. And maybe the crazyflow upgrade introduced the masking (not sure), the way we load params does not work. 

In this PR, we use the param loader from `crazyflow.drones` to load hardware params, and use` crazyflow.dynamics.core` to load dynamics related params e.g. some coefficients.

tested in real by running `tests/deploy/test_deployment.py`

### DIFF
--- a/lsy_drone_racing/control/attitude_controller.py
+++ b/lsy_drone_racing/control/attitude_controller.py
@@ -14,8 +14,7 @@ import math
 from typing import TYPE_CHECKING
 
 import numpy as np
-from crazyflow.dynamics import available_dynamics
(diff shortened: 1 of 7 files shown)


amacati (approved):
So much better, thanks a lot @Yuming-Lee24 ! Could you maybe look at https://github.com/learnsyslab/crazyflow/issues/92 as well and open a PR? Then we can make the parameter loading unified here as well once 0.3 is released

PR #128, opened 2026-08-10
Pixi run mocap runs now also nominal_frame_publisher

rducrist:
For easier handling we can now launch the `nominal_frame_publisher` from `pixi run mocap`. The new bash script launches the frame publisher as a background process in the same terminal instance. 
Further, for not having to select the topic and MarkerArray **manually**, I added my rviz config that is being loaded as well. Cheers. 

PS: The cleaner solution would've been to add the config file to the ros ws directly but since `motion_capture_tracking` is not used exclusively for drone racing I looked for a more convenient way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ scripts = ["tools/setup_mocap.sh"]
 
 ## feature [deploy]
 [tool.pixi.feature.deploy.tasks]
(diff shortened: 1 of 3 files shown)


amacati (commented):
Do we need the mocap.rviz file, i.e. does removing it mean you first have to go through all the settings?

amacati (approved):

PR #134, opened 2026-08-28
Pin MuJoCo below 3.11

### LINKED ISSUE #131: Rotating the simulation during deployment leads to crashes
We tried to rotate the camera today, and discovered that it made the drone crash immediately. Might also be a problem for pure sim, we have not checked. Can you try to replicate this and fix it? @rducrist @Yuming-Lee24

Yuming-Lee24:
Fixes #131


amacati (approved):
Can you leave a small comment that this is because the viewer breaks, and can be unlocked when gymnasium fixes the viewer?

amacati (approved):

PR #136, opened 2026-09-03
Add macOS support and update to crazyflow 0.3.2

Yuming-Lee24:
crazyflow 0.3.2 removed `device` from `SimCore`, update the affected code
only `crazyflow` is updated from 0.3.0 to 0.3.2 in the lock file by `pixi update crazyflow`

### DIFF
--- a/tools/setup_acados.sh
+++ b/tools/setup_acados.sh
@@ -13,16 +13,17 @@ if [ ! -f ${PIXI_PROJECT_ROOT}/pixi.lock ]; then
   exit 0
 fi
 
+# Set platform-specific variables so the script runs on Linux and macOS.
 case "$(uname -s)-$(uname -m)" in
   Linux-x86_64)
     LIB_EXT="so"
-    PARALLEL_JOBS="$(nproc)"
+    CPU_COUNT="$(nproc)"
     T_RENDERER_URL="https://github.com/acados/tera_renderer/releases/download/v0.0.34/t_renderer-v0.0.34-linux"
     CMAKE_PLATFORM_ARGS=()
     ;;
   Darwin-arm64)
     LIB_EXT="dylib"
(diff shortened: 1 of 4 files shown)


tools/setup_acados.sh
```diff
@@ -13,6 +13,25 @@ if [ ! -f ${PIXI_PROJECT_ROOT}/pixi.lock ]; then
   exit 0
 fi
 
+case "$(uname -s)-$(uname -m)" in
```
ratheron, line 17: We should add a comment above that here we set certain variables such that the script runs on linux and osx

tools/setup_acados.sh
```diff
@@ -13,6 +13,25 @@ if [ ! -f ${PIXI_PROJECT_ROOT}/pixi.lock ]; then
...
+  Linux-x86_64)
+    LIB_EXT="so"
+    PARALLEL_JOBS="$(nproc)"
+    T_RENDERER_URL="https://github.com/acados/tera_renderer/releases/download/v0.0.34/t_renderer-v0.0.34-linux"
+    CMAKE_PLATFORM_ARGS=()
+    ;;
+  Darwin-arm64)
+    LIB_EXT="dylib"
+    PARALLEL_JOBS="$(sysctl -n hw.ncpu)"
```
ratheron: this is the number of CPU cores, no? Maybe we should use a different variable name.

ratheron (changes requested):

tools/setup_acados.sh
Yuming-Lee24: should I change the name to CPU_COUNT both for linux and macos?

ratheron: N_CPU also works imo

ratheron (approved):

PR #137, opened 2026-09-06
support Windows

Yuming-Lee24:
support native windows, including acados setup, verified on my win11 pc, I will verify again on my laptop. The simulation seems very smooth in the renderer.

A MinGW-w64 toolchain is introduced to compile acados and MPC solver DLLs.

### DIFF
--- a/docs/getting_started/setup.md
+++ b/docs/getting_started/setup.md
@@ -49,7 +49,7 @@ Start by forking the [lsy_drone_racing](https://github.com/learnsyslab/lsy_drone
 
 If you're new to GitHub, refer to the [GitHub documentation on forking](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo).
 
-### Setting up your environment
+### Setting up your environment (Ubuntu or macOS)
 
 #### Pixi package manager (Recommended)
 
@@ -83,7 +83,6 @@ You may also use  with [micromamba](https://mamba.readthedocs.io/en/latest/insta
 
 #### Docker (Not recommended)
 
-We also provide a dev container for the simulation environment. However, this is not recommended, since it's heavy and only supports software rendering at the moment. If you're using Windows, make sure to install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install) and [Docker Desktop](https://docs.docker.com/desktop/windows/install/) with WSL integration. Refer to the `Using Docker` section below for more details.
 
 #### Simulation & Hardware on our Lab PC (If Necessary)
 
@@ -264,9 +263,8 @@ However, rendering might not work out of the box. To enable software rendering, 
 ```bash
 export LIBGL_ALWAYS_INDIRECT=1
 python scripts/sim.py -r
-```
 
-### Native Windows 10/11 (x86-64)
+### Native Windows 10/11 (x86-64, not recommended)
 
 Supports CPU simulation, MuJoCo visualization, and acados MPC without WSL or Docker.
 
@@ -339,7 +337,6 @@ For Windows, you require WSL2 to run the dev container, which enables a Linux en
     pixi shell
     ```
 
-This activates the Pixi environment with all required dependencies installed. You should now be ready to develop with the dev container!
 
 ## Common errors
 
--- a/docs/getting_started/setup.md
+++ b/docs/getting_started/setup.md
@@ -266,6 +266,49 @@ export LIBGL_ALWAYS_INDIRECT=1
 python scripts/sim.py -r
 ```
 
+### Native Windows 10/11 (x86-64)
+
+Supports CPU simulation, MuJoCo visualization, and acados MPC without WSL or Docker.
+
+**Installation and Setup**
+
+1. Install tools
+
+    Open PowerShell and check `winget --version`. If unavailable, install **App Installer**
+    from the Microsoft Store ([instructions](https://learn.microsoft.com/en-us/windows/package-manager/winget/#install-winget)). Then run:
+
+    ```powershell
+    winget install --exact --id Git.Git
+    winget install --exact --id prefix-dev.pixi
+    winget install --exact --id Microsoft.VCRedist.2015+.x64
+    ```
+
+    Download a **Win64/x86_64 GCC release ZIP (UCRT, POSIX)** from
+    [WinLibs](https://winlibs.com/#download-release). Extract `mingw64` to
+    `C:\tools\mingw64`, then add `C:\tools\mingw64\bin` to your user **Path** under
+    **Edit environment variables for your account**. Reopen PowerShell and verify:
+
+    ```powershell
+    git --version
+    pixi --version
+    gcc -dumpmachine          # Expected: x86_64-w64-mingw32
+    g++ --version
+    mingw32-make --version
+    ```
+
+    Pixi installs Python and CMake automatically. Visualization requires an OpenGL-capable graphics driver.
+
+2. Clone and install
+
+    Clone your forked course repo and install the environment:
+
+    ```powershell
+    git clone https://github.com/learnsyslab/lsy_drone_racing.git
+    cd .\lsy_drone_racing
+    pixi install
+    ```
+This activates the Pixi environment with all required dependencies installed. You should now be ready to develop with Windows.
+
 ### Dev Container (Windows 11 WSL2)
 
 For Windows, you require WSL2 to run the dev container, which enables a Linux environment within Windows. Follow these steps to set up the dev container in VS Code with WSL2:
--- a/docs/getting_started/setup.md
+++ b/docs/getting_started/setup.md
@@ -81,9 +81,6 @@ On the first invocation, Pixi will automatically resolve and install all require
 
 You may also use  with [micromamba](https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html). We do not recommend this, since Mamba/Conda environments have the tendency to be leaky and share some system-wide packages. In our experience, this will lead to problems, which is why this project is optimized to use with pixi. If you want to use micromamba anyway, we can't guarantee support.
 
-#### Docker (Not recommended)
-
-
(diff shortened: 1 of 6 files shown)


docs/getting_started/setup.md
ratheron:
```suggestion
```

I would argue we don't need the devcontainer anymore. We should also remove all files related to it. What do you thing @amacati

docs/getting_started/setup.md
```diff
@@ -266,6 +266,49 @@ export LIBGL_ALWAYS_INDIRECT=1
 python scripts/sim.py -r
 ```
```
ratheron:
```suggestion
```
This should be removed then, no?

docs/getting_started/setup.md
```diff
@@ -266,6 +266,49 @@ export LIBGL_ALWAYS_INDIRECT=1
...
 
+### Native Windows 10/11 (x86-64)
```
ratheron:
```suggestion
### Native Windows 10/11 (x86-64, not recommended)
```

docs/getting_started/setup.md
ratheron:
```suggestion
```

ratheron:
```suggestion
### Setting up your environment (Ubuntu or macOS)
```

ratheron, line 1: I think the whole file should be restructured, since we support all OS types now.

tools/setup_acados.bat
```diff
@@ -0,0 +1,100 @@
+@echo off
+rem Native Windows setup, called by Pixi through cmd.exe. No PowerShell required.
```
ratheron: Called by pixi _through cmd.exe_?

ratheron (changes requested):
Thanks for adding this. This will ease up the first week of the semester. @amacati do you agree that we should drop devcontainer support? Or just keep it in?

I've added some comments. Depending on Martin's answer, we should restructure the installation docs completely. Rest looks fine to me.

tools/setup_acados.bat
Yuming-Lee24: hmm, at first I am about to use a .ps1 to setup acados, but the pixi executes activation scripts through cmd.exe by default, so a .ps1 file can not be runned. Well this comment should be deleted.

docs/getting_started/setup.md
```diff
@@ -255,48 +251,48 @@ After that, reopen your environment. This automatically adds the package to the
...
+    winget install --exact --id Git.Git
+    winget install --exact --id prefix-dev.pixi
+    winget install --exact --id Microsoft.VCRedist.2015+.x64
+    ```
 
-    - Check the dev container configuration file `.devcontainer/devcontainer.json`.
-    - Comment out/Uncomment the necessary settings.
+    Download a **Win64/x86_64 GCC release ZIP (UCRT, POSIX)** from
+    [WinLibs](https://winlibs.com/#download-release). Extract `mingw64` to
+    `C:\tools\mingw64`, then add `C:\tools\mingw64\bin` to your user **Path** under
```
ratheron: Do we need to install this manually? Why can't we have this in pixi?

docs/getting_started/setup.md
```diff
@@ -255,48 +251,48 @@ After that, reopen your environment. This automatically adds the package to the
...
 
-2. Clone this project into the WSL2 file system (e.g., `/home/~`) rather than the Windows file system. You can access the WSL filesystem by opening a WSL2/Ubuntu terminal. Performance is significantly better when working on the WSL file system compared to the Windows file system.
+    Open PowerShell and check `winget --version`. If unavailable, install **App Installer**
+    from the Microsoft Store ([instructions](https://learn.microsoft.com/en-us/windows/package-manager/winget/#install-winget)). Then run:
 
-3. Verify dev container configuration:
+    ```powershell
+    winget install --exact --id Git.Git
+    winget install --exact --id prefix-dev.pixi
+    winget install --exact --id Microsoft.VCRedist.2015+.x64
```
ratheron: Do we need to install this manually? Why can't we have this in pixi?

docs/getting_started/setup.md
```diff
@@ -255,48 +251,48 @@ After that, reopen your environment. This automatically adds the package to the
...
+1. Install tools
 
-2. Clone this project into the WSL2 file system (e.g., `/home/~`) rather than the Windows file system. You can access the WSL filesystem by opening a WSL2/Ubuntu terminal. Performance is significantly better when working on the WSL file system compared to the Windows file system.
+    Open PowerShell and check `winget --version`. If unavailable, install **App Installer**
+    from the Microsoft Store ([instructions](https://learn.microsoft.com/en-us/windows/package-manager/winget/#install-winget)). Then run:
 
-3. Verify dev container configuration:
+    ```powershell
+    winget install --exact --id Git.Git
+    winget install --exact --id prefix-dev.pixi
```
ratheron: We should just link to the official installation websites.

docs/getting_started/setup.md
```diff
@@ -255,48 +251,48 @@ After that, reopen your environment. This automatically adds the package to the
 !!! note
     Changing the `pyproject.toml` will also update the `pixi.lock` file, which pins the exact versions of all packages. Make sure to commit both files to your repository, otherwise the tests on GitHub will fail.
 
-### Windows Subsystem for Linux (WSL2)
-
-It is also possible to develop the project on Windows using WSL2. Clone this project into the WSL2 file system, and follow the same instructions as for Linux.
-
-However, rendering might not work out of the box. To enable software rendering, set the following environment variable in your WSL2 terminal:
+### Native Windows 10/11 (x86-64, not recommended)
```
ratheron: This should be moved up after the Unix installation

ratheron (changes requested):

docs/getting_started/setup.md
```diff
@@ -215,6 +215,51 @@ python scripts/deploy.py --config level2.toml --controller <your_controller.py>
...
+    Reopen PowerShell and verify:
+
+    ```powershell
+    git --version
+    pixi --version
+    ```
+
+    Pixi installs Python, CMake, MinGW-w64 GCC/G++, GNU Make, and the Visual C++ runtime
+    automatically into the project environment from the Windows dependencies in
+    `pyproject.toml`.
```
ratheron, line 238: This is a Claude comment of what changed, but we dont need.

pyproject.toml
```diff
@@ -81,6 +81,12 @@ openblas = ">=0.3.30,<0.4"
 
 [tool.pixi.target.win-64.dependencies]
 cmake = "==3.26.0"
+vc14_runtime = "*"
+# MinGW-w64 toolchain for acados; exact package builds are pinned in pixi.lock.
```
ratheron, line 85: Exact packages should not be pinned in the lock file. The lock file is only for reproducability. If we need exact packages, pin them here.

docs/getting_started/setup.md
```diff
@@ -215,6 +215,51 @@ python scripts/deploy.py --config level2.toml --controller <your_controller.py>
 !!! note
     Be careful when flying the drone! Make sure to kill the process (**Ctrl+C**) immediately when your controller is unstable.
 
+### Native Windows 10/11 (x86-64, not recommended)
```
ratheron, line 218: Since this is properly integrated now, this can be combined with the ubuntu install above. Like, git is already listed as a prerequisite and the clone and pixi steps are identical. We should just point out what doesnt work, like, does GPU work, does deployment work (most likely not) etc. In those cases, we can add something like I propose above

docs/getting_started/setup.md
ratheron, line 153: #### Install deployment environment (deploy controller to real drones, Ubuntu only)

If this is the case.

ratheron (changes requested):
Nice that all dependencies are in pixi now. This makes it super easy to install. Thanks!

Since in this case the tutorial is basically identical for all platforms, we should only have one tutorial, which is just pointing out where differences lie (like file structure in windows etc)

docs/getting_started/setup.md
```diff
@@ -81,10 +81,6 @@ On the first invocation, Pixi will automatically resolve and install all require
 
 You may also use  with [micromamba](https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html). We do not recommend this, since Mamba/Conda environments have the tendency to be leaky and share some system-wide packages. In our experience, this will lead to problems, which is why this project is optimized to use with pixi. If you want to use micromamba anyway, we can't guarantee support.
 
-#### Docker (Not recommended)
-
-We also provide a dev container for the simulation environment. However, this is not recommended, since it's heavy and only supports software rendering at the moment. If you're using Windows, make sure to install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install) and [Docker Desktop](https://docs.docker.com/desktop/windows/install/) with WSL integration. Refer to the `Using Docker` section below for more details.
-
 #### Simulation & Hardware on our Lab PC (If Necessary)
```
Yuming-Lee24: why this is If nessary? don't we prefer students to deploy on lab PCs? @ratheron

ratheron: I agree, the `(If Necessary)` can be removed imo

docs/getting_started/setup.md
Yuming-Lee24: these two lines, a robostack env is not something that student need to install before setting up the repo env, the robotstack is installed via `pixi install -e deploy`, I suggest to remove this line. And also the line for docker, since we don't support it anymore I guess? @ratheron

Yuming-Lee24: not a fan of putting these here, confusing: "should I execute these commands or not?", I would cover this in the simulation env setup below

Yuming-Lee24: we don't need -r now, rendering option is in the config file

ratheron: Docker, yes. Robostack, we should still hint what the env is

Yuming-Lee24: @ratheron

Yuming-Lee24: @ratheron

